PR #22, opened 2018-08-01
fix(merge): merge multiple times without errors

### LINKED ISSUE #21: Duplicate << keys shouldn't be marked as error when merge=true
version: 1.0.0-rc.7

```js
var YAML = require("yaml").default;

YAML.parse(`
.anchors:
  - &anchor1
    key: value
  - &anchor2
    another: prop
foo:
  bar: baz
  <<: *anchor1
  <<: *anchor2
`, { merge: true }) //=> YAMLSemanticError: Map keys must be unique; "<<" is repeated
```

Expected: no error

---

Original post: prettier/prettier#4919

ikatyang:
Fixes #21

Skip duplicate key check for merge key.

### DIFF
--- a/__tests__/anchors.js
+++ b/__tests__/anchors.js
@@ -141,6 +141,16 @@ describe('merge <<', () => {
     expect(String(doc)).toBe('[ &a1 { a: A }, { b: B, <<: *a1 } ]\n')
   })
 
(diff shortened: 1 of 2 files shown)


eemeli:
That looks like just the right fix for this, thank you!

PR #25, opened 2018-08-02
feat: include offset in FlowChar

ikatyang:
Extracted from #24

change `FlowChar` from `string` to `{ char: string, offset: number }`

### DIFF
--- a/__tests__/cst/common.js
+++ b/__tests__/cst/common.js
@@ -29,13 +29,13 @@ export const pretty = node => {
 export const testSpec = (res, exp) => {
   if (typeof exp === 'string') {
     const value =
-      res.char
-        ? res.char
-        : res instanceof Node
-          ? 'strValue' in res
-            ? res.strValue
-            : res.rawValue
-          : res
+      res instanceof Node
+        ? 'strValue' in res
+          ? res.strValue
+          : res.rawValue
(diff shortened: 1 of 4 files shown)


__tests__/cst/common.js
```diff
@@ -33,7 +33,9 @@ export const testSpec = (res, exp) => {
         ? 'strValue' in res
           ? res.strValue
           : res.rawValue
-        : res
+        : typeof res === 'string'
+          ? res
+          : res.char
```
ikatyang: It's possible to be string here:

```
  ● 7.3.3. Plain Style › Example 7.10. Plain Characters

    expect(received).toBe(expected) // Object.is equality

    Expected: " Outside flow collection:"
    Received: undefined

    Difference:

      Comparing two different types of values. Expected string but received undefined.

      35 |           : res.rawValue
      36 |         : res.char
    > 37 |     expect(value).toBe(exp)
         |                   ^
      38 |   } else if (Array.isArray(exp)) {
      39 |     expect(res).toBeInstanceOf(Array)
      40 |     trace: 'test-array', exp
```

eemeli: Ah, you're, right, I'd forgotten how `testSpec` is used by the tests for comment values.

PR #29, opened 2018-08-04
fix(merge): no error when key is null

ikatyang:
`iKey` is possible to be `null`.


ikatyang:
Actually, I didn't do so since I'm not sure where to put it, or should I add a new test file?

eemeli:
Should probably clean up at least the test file names at some point. `corner-cases.js` has turned into a bit of a catch-all, but `anchors.js` works as well for this.

PR #45, opened 2018-11-15
fix(setOrigRanges): add missing props

ikatyang:
- there's no `valueRange` in comments
- fake node should respect `setOrigRanges`


eemeli:
Looks good. Do you have any other fixes lined up in the near future, or should I just release this as 1.0.2?

ikatyang:
Thanks for the quick response!

This should be all the bugs related to CRLF. After this fix, my tests are passed and I'm able to support CRLF in yaml-unist-parser. ❤️

PR #51, opened 2018-11-28
Add option to output JS Maps, add generic support for iterables

### LINKED ISSUE #46: Using custom tags as keys?
Thanks for this library, this looks awesome and I'm going to play with it a bit.

Can we use custom tags as keys? i.e.:

```yaml
foo:
    !bar: Hello
```

### LINKED ISSUE #49: Support stringifying JavaScript Map as !!map
A [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) should obviously be a `!!map`.

Following discussion in #46

### LINKED ISSUE #50: Add option to parse !!map as a Map rather than an Object
Based on discussion in #46. In JS `Object` keys are forced to string, while `Map` keys are not, and both are arguably good representations of a YAML `!!map`.

This also requires stringifier support, that's in issue #49.

eemeli:
Fixes #46, fixes #49, fixes #50 

On the YAML -> JS side, this adds a new option `mapAsMap`, with which YAML mappings are represented in JS as `Map` rather than `Object` instances:

```js
YAML.parse('{[1, 2]: many}')
// { '[1,2]': 'many' }

YAML.parse('{[1, 2]: many}', { mapAsMap: true })
// Map { [ 1, 2 ] => 'many' }
```

The specific benefit of this is the ability to natively use non-string keys in mappings.

Then on the JS -> YAML side, iterable values are now supported by `createNode()`. The default is to parse them as sequences, but `Map` and its descendants are parsed as mappings.

### DIFF
--- a/__tests__/createNode.js
+++ b/__tests__/createNode.js
@@ -169,3 +169,47 @@ z:
     })
   })
 })
+
+describe('Set', () => {
+  test('createNode(new Set)', () => {
+    const s = createNode(new Set())
+    expect(s).toBeInstanceOf(Seq)
+    expect(s.items).toHaveLength(0)
+  })
+  test('createNode(new Set([true]), false)', () => {
+    const s = createNode(new Set([true]), false)
+    expect(s).toBeInstanceOf(Seq)
+    expect(s.items).toMatchObject([true])
+  })
+  describe('new Set([3, new Set(["four", 5])])', () => {
+    const set = new Set([3, new Set(['four', 5])])
+    test('createNode(value, false)', () => {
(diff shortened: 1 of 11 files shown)


eemeli:
I'm considering adding a detector for Immutable.js maps as well, with something like `iterable.constructor.isMap && iterable.constructor.isMap(iterable)` as an alternative for `iterable instanceof Map`. Any thoughts on whether that would make sense, or would it be too specific? Checking `iterable[Symbol.iterator].name === 'entries'` might work as well, but it's a bit flaky.

PR #54, opened 2018-11-28
Stop including end comments in collections if next content is not sufficiently indented

### LINKED ISSUE #41: Indentation level is not taken into account for trailing comments
This is split off from #28, which was reported by @douglasmuraoka.

With input like this:
```
a1:
a2:
  b1: c1
  b2: c2
#comment
```
the comment is attached to the b-level collection, rather than the root a-level collection. This means that when re-stringifying, the comment will be indented as well.

To fix, the indentation level of such trailing comments should be taken into account when determining which collection to attach them to.

eemeli:
Fixes #41 

Effectively, when encountering a comment on a new line while parsing a collection, the parser now peeks ahead to see if the next non-comment content is indented at least to the level of this collection, or if there's a subsequent comment that's sufficiently indented. This means that e.g. parsing and re-stringifying this document:

```yaml
#c0
- #c1
  k1: v1
  #c2
  k2: v2 #c3
#c4
  k3: v3
#c5
```

Results in this representation:

```yaml
#c0
#c1
- k1: v1
  #c2
  k2: v2 #c3
  #c4
  k3: v3
#c5
```

Previously, the `#c5` comment would have gained a spurious level of indentation due to being incorrectly attached to the `k*` map rather than the root sequence.

@ikatyang Did you have some fixes in `yaml-unist-parser` for this following prettier/prettier#4917 that would need to take into account these changes?

### DIFF
--- a/__tests__/comments.js
+++ b/__tests__/comments.js
@@ -114,7 +114,15 @@ describe('parse comments', () => {
       expect(doc.contents.items[0].items[1].commentBefore).toBe('c2')
       expect(doc.contents.items[0].items[1].value.comment).toBe('c3')
       expect(doc.contents.items[0].items[2].commentBefore).toBe('c4')
-      expect(doc.contents.items[0].comment).toBe('c5')
+      expect(doc.contents.comment).toBe('c5')
+      expect(String(doc)).toBe(`#c0
+#c1
+- k1: v1
+  #c2
+  k2: v2 #c3
+  #c4
+  k3: v3
+#c5\n`)
     })
   })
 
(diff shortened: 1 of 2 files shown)


ikatyang:
Thanks for the notice. Basically, `yaml-unist-parser` re-attaches every comment (with a few exceptions) so the result should be the same as long as all the comments are collected correctly. I just tried this branch in `yaml-unist-parser`, all the tests are passed.

I think this kind of comment attaching can be hard for a parser since it needs to be fixed one-by-one, so I didn't raise this sort of issues here. Instead, I apply a postprocessing for comment attaching in `yaml-unist-parser`, which I just need to define general rules (something like _"when A comment is in front of B node, attach A to B"_) and it'll work for all the cases.

PR #64, opened 2018-12-24
Add blank lines

### LINKED ISSUE #42: Empty lines should not be discarded
This is split off from #28, which was reported by @douglasmuraoka.

Empty lines can help semantically separate blocks of content, and so they should not be discarded as is currently done.

eemeli:
Note: This is still a work in progress, and the API described here may still develop a bit. When ready, this fixes #42.

At the **CST** level this adds a `BLANK_LINE` node type, which may represent multiple consecutive empty lines. The parsing of `range.end` and for plain values `valueRange.end` needed to get adjusted/corrected a bit to allow for this.

As a special case, empty lines following block scalar values with "keep" chomping (i.e. with `+` in their header) are considered to be a part of the node value, and therefore won't be parsed as `BLANK_LINE` nodes.

**AST** nodes get new `spaceBefore` boolean attributes that represent empty lines. Multiple empty lines are collapsed into one, and empty lines between a node and its preceding comment are dropped. When stringifying, the order in which a node is represented is:

1. `spaceBefore`
2. `commentBefore`
3. `value`
4. `comment`

Comments at the very beginning that are separated from the rest of the contents with a blank line are captured in the document's `commentBefore` value; similarly top-level comments at the very end are included in its `comment`. When stringifying, these document comments will be separated from the contents where possible. Blank lines at the very beginning and end of the document are trimmed when parsing.

~I'm considering dropping `spaceAfter` as unnecessary (and as of yet only partly implemented), but at least for now the parsing algorithm is pretty greedy, and grabs trailing blank lines into the innermost collection that they could belong to, which effectively requires `spaceAfter` in the AST.~

**Edit:** `spaceAfter` is now gone, and blank lines & less-indented comments at collection end now get assigned to the parent collection.

**Edit 2:** Document comments now get automatic spaces, and the document no longer has a `spaceBefore` property.

### Todo
- [x] Parse blank lines within flow collections
- [x] Attach blank lines more appropriately to following rather than preceding nodes
- [x] Either drop `spaceAfter` or make sure it always works
- [x] Add moar tests
- [x] Fix document comments & their spacing


eemeli:
Edited the top comment to reflect updates. In addition to what's mentioned there, the parsing of comments within map pairs has been updated a bit: the `Pair` object itself can now have its own comment, rather than aliasing that of its value. This is a comment after the key & indicator, but before the value. In other words, you now get this representation:
```js
const doc = YAML.parseDocument(`
#0
key: #1
  #2
  value #3`)

> pair = doc.contents.items[0]
Pair {
  key:
   Scalar {
     value: 'key',
     range: [ 4, 7 ],
     type: 'PLAIN',
     commentBefore: '0' },
  value:
   Scalar {
     value: 'value',
     range: [ 19, 27 ],
     type: 'PLAIN',
     commentBefore: '2',
     comment: '3' },
  type: 'PAIR',
  comment: '1' }
```

Assigning a multi-line comment to `pair.comment` will stringify those on separate lines, which means that on subsequent parses they'll get picked up as a part of `pair.value.commentBefore`.

eemeli:
I think this might be about done now? I may let this sit open for a little while before merging, in case I change my mind.

Issue #67, opened 2019-01-04
Range in characters is sub-optimal for error reporting to humans


MikeRalphson:
As YAML is line-delimited (can't be strung together minified like JSON) it makes at least some sense to be able to report errors (both syntactically and semantically) as a line number and column number pair.

Is it simple to add a helper routine to achieve this? Happy to PR with a bit of pointing in the right direction.

My end use-case is to turn JSON pointer references to schema violations back into line number, column number pairs.

eemeli:
That sounds like a good idea, and a PR would indeed be welcome. This may need to expose two different APIs, e.g. something like `charPosToLineCol(pos, str): { line: number, col: number }` for the actual implementation (you may be able to find something already on npm) and a wrapper convenience method added to the errors (which due to this may need a common ancestor).

Things to be aware of:
- A `Document` or its parsed nodes do not directly contain any reference to the source string, unless `keepCstNodes` is [enabled](https://eemeli.org/yaml/#options).
- Parsed AST nodes contain a `range: [start, end]` property; nodes not generated by parsing obviously will not.
- CST nodes always have a `range: { start: number, end: number }` property
- Errors always contain a `source` reference to their CST node, which in turn do contain a reference to the source string as `context.src`
- When parsing, if the input contains `\r\n` line terminators, those are converted to `\n` terminators before processing. The string at `context.src` will only have `\n` line terminators. The CST parser's [ability](https://eemeli.org/yaml/#cst-setorigranges-bool) to add `origRange` pointers to the real original source isn't (currently) available when using the library via e.g. `YAML.parse()` or `YAML.parseDocument()`, and should not be strictly required for this.

MikeRalphson:
Ok thanks. I'm making some progress, able to build up a list of line offsets at the same point the `\r`s are removed in the CST parser which should allow a new `charPosToLineCol` function to be performant.

Will PR as soon as I have a tested solution. Please consider me working on this issue.

eemeli:
Correct. The parser doesn't otherwise need to track newlines, so we should not require that it does so every time. However, as you may get multiple errors from a single input document, it may well make sense to cache the positions of newlines when they're first determined.

MikeRalphson:
Thanks. Will put the calculation behind an option.

MikeRalphson:
Perfect, thanks.

MikeRalphson:
```
mike:~/nodejs/yaml$ git submodule update --init
[...]

PR #68, opened 2019-01-05
Optionally compute lineOffsets, refs #67

MikeRalphson:
As discussed in #67 this PR adds the optional ability to compute line offsets and provides a helper function `charPosToLineCol`.

Both the AST and CST parse(Document) functions can now accept a `computeLineOffsets` option. Tests added to ensure default behaviour is unchanged and for the new functionality.

`README.md` has been minimally updated, but not yet the docs. I'd appreciate some guidance on doing that.

### DIFF
--- a/src/cst/parse.js
+++ b/src/cst/parse.js
@@ -3,14 +3,22 @@
 import Document from './Document'
 import ParseContext from './ParseContext'
 
-export default function parse(src) {
+export default function parse(src, options) {
+  const defaultOptions = { computeLineOffsets: false }
+  options = Object.assign({}, defaultOptions, options)
   const cr = []
+  const lf = [0]
   if (src.indexOf('\r') !== -1) {
     src = src.replace(/\r\n?/g, (match, offset) => {
       if (match.length > 1) cr.push(offset)
       return '\n'
     })
   }
+  if (options.computeLineOffsets)
+    src = src.replace(/\n/g, (match, offset) => {
+      lf.push(offset + 1)
+      return '\n'
+    })
   const context = new ParseContext({ src })
   const documents = []
   let offset = 0
@@ -30,5 +38,25 @@ export default function parse(src) {
     return true
   }
   documents.toString = () => documents.join('...\n')
+  if (options.computeLineOffsets) documents.lineOffsets = lf
   return documents
+}
+
+export function charPosToLineCol(offset, lineOffsets) {
+  if (
+    !lineOffsets ||
+    typeof offset === 'undefined' ||
+    offset < 0 ||
+    offset > lineOffsets[lineOffsets.length - 1]
+  )
+    return { line: undefined, col: undefined }
+  const lineIndex = lineOffsets.indexOf(offset)
+  if (lineIndex >= 0)
+    return { line: lineIndex, col: offset - lineOffsets[lineIndex] }
+  for (let i = 0; i < lineOffsets.length; i++) {
+    if (lineOffsets[i] > offset) {
+      return { line: i - 1, col: offset - lineOffsets[i - 1] }
+    }
+  }
+  return { line: undefined, col: undefined }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,8 @@ const defaultOptions = {
   keepBlobsInJSON: true,
   mapAsMap: false,
   tags: null,
-  version: '1.2'
+  version: '1.2',
+  computeLineOffsets: false
 }
 
 function createNode(value, wrapScalars = true, tag) {
@@ -34,12 +35,17 @@ class Document extends YAMLDocument {
 }
 
 function parseAllDocuments(src, options) {
-  return parseCST(src).map(cstDoc => new Document(options).parse(cstDoc))
+  return parseCST(src, {
+    computeLineOffsets: options ? options.computeLineOffsets : false
+  }).map(cstDoc => new Document(options).parse(cstDoc))
 }
 
 function parseDocument(src, options) {
-  const cst = parseCST(src)
+  const cst = parseCST(src, {
+    computeLineOffsets: options ? options.computeLineOffsets : false
+  })
   const doc = new Document(options).parse(cst[0])
+  if (options && options.computeLineOffsets) doc.lineOffsets = cst.lineOffsets
   if (cst.length > 1) {
     const errMsg =
       'Source contains multiple documents; please use YAML.parseAllDocuments()'
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ import Seq from 'yaml/seq'
 import parseCST from 'yaml/parse-cst'
 ```
 
-* [`parseCST(str): CSTDocument[]`](https://eemeli.org/yaml/#parsecst)
-* [`YAML.parseCST(str): CSTDocument[]`](https://eemeli.org/yaml/#parsecst)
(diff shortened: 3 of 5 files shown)


src/cst/parse.js
```diff
@@ -30,5 +38,25 @@ export default function parse(src) {
...
+  if (
+    !lineOffsets ||
+    typeof offset === 'undefined' ||
+    offset < 0 ||
+    offset > lineOffsets[lineOffsets.length - 1]
+  )
+    return { line: undefined, col: undefined }
+  const lineIndex = lineOffsets.indexOf(offset)
+  if (lineIndex >= 0)
+    return { line: lineIndex, col: offset - lineOffsets[lineIndex] }
```
eemeli: This is special-case handling for when `offset` points to a `\n` character, yes? Why couldn't it be handled by the following `for` loop?

src/cst/parse.js
```diff
@@ -3,14 +3,22 @@
...
     src = src.replace(/\r\n?/g, (match, offset) => {
       if (match.length > 1) cr.push(offset)
       return '\n'
     })
   }
+  if (options.computeLineOffsets)
+    src = src.replace(/\n/g, (match, offset) => {
+      lf.push(offset + 1)
+      return '\n'
+    })
```
eemeli: `src.replace()` is far too heavy an operation for this; use [`src.indexOf('\n', fromIndex)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) instead. There's no reason to mutate the string here.

src/index.js
```diff
@@ -34,12 +35,17 @@ class Document extends YAMLDocument {
 }
 
 function parseAllDocuments(src, options) {
-  return parseCST(src).map(cstDoc => new Document(options).parse(cstDoc))
+  return parseCST(src, {
+    computeLineOffsets: options ? options.computeLineOffsets : false
+  }).map(cstDoc => new Document(options).parse(cstDoc))
```
eemeli: The result of the `cst.map()` operation will not contain the non-iterable properties of `cst`, which means that setting `computeLineOffsets` will not influence the output of this function.

src/cst/parse.js
```diff
@@ -30,5 +38,25 @@ export default function parse(src) {
     return true
   }
   documents.toString = () => documents.join('...\n')
+  if (options.computeLineOffsets) documents.lineOffsets = lf
   return documents
 }
+
+export function charPosToLineCol(offset, lineOffsets) {
```
eemeli: Is there a particular reason why this function is being exported from `cst/parse.js`, rather than its own file?

src/cst/parse.js
MikeRalphson: I may have optimised pre-emptively here (which is a mistake without measuring it). It just avoids the search-in-Javascript and allows the optimised `indexOf` method to have a chance. As the array is always sorted, perhaps a binary search-like approach would be better than this plus the loop, especially for large documents?

src/cst/parse.js
MikeRalphson: Cool. Will amend in the next revision.

src/index.js
MikeRalphson: Ok, will take another look.

src/cst/parse.js
MikeRalphson: No, just was unsure of your convention on this.

MikeRalphson:
> I don't have a complete solution envisioned here ... What do you think? Sorry to leave this response a bit incomplete; I'll need to return to this a bit later.

I've got no problems with computing the `lineOffsets` on the first call to `charPosToLineCol`, we can use the presence or absence of `cst.lineOffsets` to do that if you're happy with a property appearing post-construction of the CST?

Would that (plus the above changes) be enough for me to do the next revision of this PR?

src/cst/parse.js
eemeli: I'd recommend first going with just a simple for loop, and then seeing if there's a real-world case that would show a need for a more complicated search pattern.

eemeli:
I'm fine with `lineOffsets` getting added later. Did you have something in mind yet for locating errors, when encountered during `YAML.parseDocument()` or `YAML.parse()`, which don't (at least for now) provide easy accesss to the root `cst` object?

MikeRalphson:
> Did you have something in mind yet for locating errors, when encountered during YAML.parseDocument() or YAML.parse(), which don't (at least for now) provide easy accesss to the root cst object?

The likely (initial) recommendation I think would be to reparse the document(s) at the CST level. I don't know how acceptable you find that... We might be able to do something for the AST where `keepCstNodes` has been specified. When you say "for now", is access to the CST from higher levels likely to change?

I plan to follow this up with another PR (or a plugin, or a separate module) which can decorate the CST with JSON pointers allowing a lookup from the JS object to the CST, and then to the line/column position. JSON pointers being widely used in the JSON-Schema and OpenAPI worlds.

Just to make you aware, the other PR/plugin/module I have in mind is the optional ability to be able to preserve comments when `parse`ing to, and `stringify`ing from a JS object, in a configurable property key.

eemeli:
If there's a good reason to extend the API, then we can do that -- but it does require a good reason. If your application is usually interested in the CST, then it makes sense to explicitly build the `cst`, and then use `new YAML.Document(options).parse(cst)` rather than `YAML.parseDocument()` to build the document -- and thereby maintain your own reference to the CST.

The main use case that I'd expect for line/col references outside of those cases would be errors, when such are not expected. It would be useful, when encountering one, to be able to use the error's context to build a line/col reference to the source. Then, if the offsets are cached as `cst.lineOffsets`, it may make sense to add a `context.root` ref to each CST Node in order to access that.

I've never really worked with JSON pointers, but I'd suspect that it'd be relatively simple to write a getter method for the AST Document that would help with that. Maybe something like `Document#get(path: Iterable): Node?`, so you could (separately) parse a JSON pointer like `"/foo/0"` into `['foo', 0]` and use that to fetch the matching Node?

Regarding comments, what use case did you have in mind for the object with comments? My suspicion is that once you account for all the comments that YAML allows for, you'd end up with something very similar to the existing Document object.

MikeRalphson:
> Regarding comments, what use case did you have in mind for the object with comments? 

Simply for passing into existing code which can only handle native JS objects in a fixed representation, and receiving it back (possibly mutated) and then writing it back out as YAML.

For example, a routine which converts an OpenAPI 2.0 document to OpenAPI 3.x. The comments could be converted into `x-comment` properties (and I understand there's complexity aound `commentBefore` etc) and then back into YAML comments on stringification. Hence possibly being a plugin or separate module rather than adding it to core.

eemeli:
@MikeRalphson I updated your branch with the changes I'd requested, so that this can be merged and released along with other recent changes.

Essentially, the public API is now a new getter `rangeAsLinePos` that's available on all CST nodes, along with a `context.root` reference for the current document. The latter is used by the former in order to cache the line start indices at `cstdoc.lineStarts`.

The `getLinePos(offset, cst)` function itself will be available at `yaml/dist/cst/getLinePos`, though that probably doesn't need to be documented.

For errors, the line positions are now available from `err.source.rangeAsLinePos`.

eemeli:
Just as a heads-up, I realised after releasing 1.3.0 that zero-indexing the line/col positions was surprising, so pushed out a patch release 1.3.1 that fixes our behaviour to follow the norm of one-indexing these positions.

PR #70, opened 2019-01-05
travis; test on Node.js LTS version also

MikeRalphson:
in addiion to 'current' and 6.x. Currently this means Node.js 10.x (10.15.0 at time of writing) will be tested as well as the current (Node.js 11) and Node.js 6 (presumably the baseline for this project).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - 'node'
+  - "lts/*"
   - '6'
 
 script:


eemeli:
Thanks, also testing the LTS release makes sense. Node 6 is indeed the current target of the Babel build as maintaining that has been relatively easy. On a related note, it might make sense to look into how `yaml` actually runs in browser environment. It should be fine, but I've not really tested it for a while.

PR #73, opened 2019-01-11
 Add a separate build for browsers

### LINKED ISSUE #72: Classes in bundled code breaks IE11 compatibility
I've been having a real hard time debugging an issue with IE11 and it turns out the offending party is `yaml`, since your babel config apparently doesn't transform classes. Is there a special reason for this? I was surprised since it's the first dependency I've come across with this issue. 80% of our customers use IE11 (yeah, I know 🙄) so It's hard to ignore.

Anyway, I'll just write out `yaml` for now. Thank you for an otherwise great library.

eemeli:
This fixes #72 by adding a separate build of the library that targets browser environments, using a browserslist rule `> 0.5%, not dead` which effectively means being limited by what IE 11 requires. At least for now this build does depend on `@babel/runtime`, which is _not_ included in the package's dependencies, as it's not needed by the Node build. Is there a way of specifying separate dependencies for the `"main"` and `"browser"` versions of a package?

It seems to work, but I'd like to verify that by adding browser-based tests. And with that I'd really appreciate some help, as I've no prior experience in setting up a CI pipeline that would allow for e.g. Jest tests to be run on Selenium in the Sauce Labs or Browserstack clouds. Anyone?

At the same time I've tuned the Node build's Babel preset-env config to target Node 6.5 in order to trim down it's somewhat overly-cautious rules, while making sure that actual compatibility with at least Node 6.0.0 is still maintained.

Based on some testing with Webpack, it looks like the browser version gzips to about 22kB (including `@babel/runtime`), but the autogenerated set of polyfills `browser/dist/polyfill.js` would add about 14kB on top of that (compressed). If we do get browser tests set up, that could/should get whittled down quite a bit.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,6 +1,6 @@
 const isBrowser = process.env.BABEL_ENV === 'browser'
 
-const envOptions = { useBuiltIns: false }
+const envOptions = { useBuiltIns: 'entry' }
 if (!isBrowser) envOptions.targets = { node: '6.5' }
 const presets = [['@babel/env', envOptions]]
 
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,8 +1,14 @@
-module.exports = {
-  presets: [['@babel/env', { targets: { node: '6.5' } }]],
-  plugins: [
-    '@babel/plugin-proposal-class-properties',
-    ['babel-plugin-trace', { strip: true }],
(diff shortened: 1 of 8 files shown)


eemeli:
I'm merging this in order to push out a new release, and will continue developing its browser tests in a [separate repo](https://github.com/eemeli/yaml-playground) as it doesn't really make sense to include all of the Webpack toolchain and other related artifacts here.

Issue #80, opened 2019-01-28
Default vs match for custom scalars


isaacs:
I'm currently using this as my Symbol type:

```js
const stringType = require('yaml/dist/schema/_string.js')
module.exports = {
  class: Symbol,
  tag: 'tag:node-tap.org,2019:symbol',
  resolve (doc, $1) {
    return Symbol($1)
  },
  test: /^Symbol\((.*)\)$/,
  stringify (item, ctx, onComment, onChompKeep) {
    return stringType.stringify(item, ctx, onComment, onChompKeep)
  },
}
```

I notice that the results are always double-quoted:

```js
{
  a: Symbol(a),
  b: 'Symbol(b)'  // not actually symbol, just matches the regexp
}
```

even though this seems excessive in the output:

```yaml
---
a: !<tag:node-tap.org,2019:symbol> "Symbol(a)"
b: "Symbol(b)"
```

This is the line doing it: https://github.com/eemeli/yaml/blob/master/src/schema/_string.js#L253-L255

If I understand this bit correctly, it's handling cases where a value would be ambiguous if unquoted, like `true`.  However, this seems like it would only be a problem if the type is established as a default; as it is, with `default: true`, Symbol objects will always be stringified with an explicit tag.

However, when I pass the `default` value through on the resolved Scalar object and let non-default values pass through unquoted, it ends up parsing as a Symbol, even without the tag.

1. Is there a way to say "no, really, this type _must_ have an explicit tag, even if the regexp matches"?
2. If such a flag is set, would it be appropriate to stringify without quoting?  And is it intelligible on parsing, that such a flag would mean "only use this type if it has that tag"?

Please let me know if I'm missing something, or if you think it's a good idea.  Thanks!

isaacs:
Note: that also means I can't just do `!<tag:node-tap.org,2019:symbol> a` which sort of seems like a cuter representation.  But if the `test` in my type is `.*`, then anything _might_ be a symbol, so every string everywhere is double-quoted.

isaacs:
Aha!  Thanks, I had a feeling I was missing something obvious.

This works exactly as I want:

```js
const stringType = require('yaml/dist/schema/_string.js')
module.exports = {
  class: Symbol,
  tag: 'tag:node-tap.org,2019:symbol',
  resolve (doc, item) {
    const src = stringType.resolve(doc, item)
    return Symbol(src)
  },
  stringify (item, ctx, onComment, onChompKeep) {
[...]

PR #82, opened 2019-01-30
set reqStringParse for all (and only) strings

isaacs:
This is another approach for the issue being discussed in #80.

Set reqStringParse for strings, but never for non-strings.


isaacs:
I think this is superseded by other changes on master now that meet the same need more appropriately. Closing. Thanks :)

PR #84, opened 2019-02-05
Handle circular references

### LINKED ISSUE #81: Support circular objects (using Anchors?)
Consider this js:

```js
const yaml = require('yaml')
const a = {}
a.a = a
console.log(yaml.stringify(a))
```

The result is a stack overflow, which is not very good.  I'd really like to avoid having to work around this by sanitizing objects prior to sending to the yaml stringifier.

One solution would be to keep a list of all the objects visited earlier in a traversal, and print `[Circular]` or something.

Another would be to take advantage of [yaml anchors](https://sourceforge.net/p/yaml/mailman/message/10965262/), so the above object would stringify as:

```yaml
---
&generated_anchor
a: *generated_anchor
...
```

or something?

You probably _don't_ want to litter the output with anchors on every map and sequence, so maybe it should only do this if it sees that an object is repeated and walks up to the parent instance to add the anchor where needed.

eemeli:
Fixes #81

Essentially, this now works:
```js
const a = ['b']
a.push(a)
const str = YAML.stringify(a)
// &a1
// - b
// - *a1
YAML.parse(str)
// [ 'b', [Circular] ]
```

For stringifying, this means that input that would earlier throw now doesn't, producing valid YAML instead. When parsing, alias nodes referring to anchors now resolve the original objects, rather than separate instances. I'm not sure yet whether this should mean a major version bump or not.

The public `YAML.createNode()` and the `doc.schema.createNode()` APIs have not changed, and do not support the autogeneration of alias nodes.

### DIFF
--- a/src/Anchors.js
+++ b/src/Anchors.js
@@ -11,6 +11,10 @@ export default class Anchors {
 
   map = {}
 
+  constructor(prefix) {
+    this.prefix = prefix
+  }
+
   createAlias(node, name) {
     this.setAnchor(node, name)
     return new Alias(node)
@@ -39,6 +43,7 @@ export default class Anchors {
   }
 
   newName(prefix) {
+    if (!prefix) prefix = this.prefix
     const names = Object.keys(this.map)
     for (let i = 1; true; ++i) {
       const name = `${prefix}${i}`
@@ -79,7 +84,7 @@ export default class Anchors {
     } else {
       if (!name) {
         if (!node) return null
-        name = this.newName('a')
+        name = this.newName()
       }
       map[name] = node
     }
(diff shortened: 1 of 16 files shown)


isaacs:
Fantastic!  I'll play with this tonight and provide some feedback.  Thanks!

isaacs:
As far as I can tell, this works perfectly.

My only complaint is that there doesn't appear to be a way to disable the warning from being printed to stderr.

```js
// dist/Document.js:455
if (!src.resolved) {
  const msg = 'Alias node contains a circular reference, which cannot be resolved as JSON';
  this.warnings.push(new _errors.YAMLWarning(node, msg));
}

// dist/index.js:66
doc.warnings.forEach(warning => console.warn(warning));
```

If it handles circular references fine, then so what?  If I'm using YAML to stringify and parse, who cares that it doesn't make for valid JSON?

eemeli:
You're right, that warning doesn't make sense any more, so I removed it.

Issue #77, opened 2019-01-26
Support for JS Symbols


isaacs:
It would be nice if this module could either handle JavaScript Symbol objects, or provide some facility to define custom types such that one could add support in userland.

js-yaml has support for custom types, but they rely on providing a class, and `instanceof` returning true.  However, `Symbol('x') instanceof Symbol` is always false (even though `Object.getPrototypeOf(Symbol('x')) === Symbol.prototype`, because sometimes JS is wat!?)  I could wrap them in a SymbolWrap class and unwrap on parse, but ugh.

Context: I'm working on refactoring the yaml handling in node-tap, and one common complaint is that Symbol objects cause a throw.

isaacs:
With that change, this seems to work sort of: https://gist.github.com/isaacs/d79a73009de66385075cf0f66093f53e

Any feedback would be welcome.  Thanks!

eemeli:
That's very wat. I think the right thing to do here would be to extend the check as you suggested. Any particular reason why you'd prefer the `Object.getPrototypeOf` check instead of e.g. this?

```js
const match = this.tags.filter(
  t =>
    t.class &&
    (obj instanceof t.class || (obj && obj.constructor === t.class))
```

Slightly longer-term, I've been considering replacing the tag `class` properties with a boolean-returning function, which would allow for more generically matching JS objects or arrays to custom tags. Hence why I'd rather apply a quick-fix for now, rather than extending the tag object spec.

For the other primary types, the `typeof` checks effectively get subsumed into the `class` checks by the immediately preceding switch case, but for Symbol I'd just assumed that it'd work in a sensible manner.

isaacs:
Yeah, `obj.constructor` is totally a fine way to go.  No reason to pick one over the other, imo.

> considering replacing the tag class properties with a boolean-returning function

That seems extremely sensible.  Then you can also put whatever other logic in there that makes sense, like sniffing at fields or string formatting.

> for Symbol I'd just assumed that it'd work in a sensible manner
That'll teach ya ;P

isaacs:
I think #79 addresses this well enough for now.

PR #86, opened 2019-02-08
Make custom tags easier

eemeli:
This is another continuation of #77, #79, and #80, and replaces some of the solutions previously implemented. Specifically, this PR is doing three things:

1. Deprecate the `class` attribute of tags, replacing it with the more generic `identify(value)`.
2. Turn the string stringifier into a generic scalar stringifier, fallback to it by default, and export it as `stringify()` from `yaml/schema`.
3. Improve the documentation on writing custom tags.

In total, this means that e.g. RegExp and shared Symbol tags could now be implemented as:

```js
import { stringify } from 'yaml/schema'

const regexp = {
  identify: value => value instanceof RegExp,
  tag: '!re',
  resolve(doc, cst) {
    const match = cst.strValue.match(/^\/([\s\S]+)\/([gimuy]*)$/)
    return new RegExp(match[1], match[2])
  }
}

const sharedSymbol = {
  identify: value => value.constructor === Symbol,
  tag: '!symbol/shared',
  resolve: (doc, cst) => Symbol.for(cst.strValue),
  stringify(item, ctx, onComment, onChompKeep) {
    const key = Symbol.keyFor(item.value)
    if (key === undefined)
      throw new Error('Only shared symbols are supported')
    return stringify({ value: key }, ctx, onComment, onChompKeep)
  }
}
```


eemeli:
Rebased on top of `master`.

@isaacs Note that this PR removes the `obj.constructor === t.class` check that #79 introduced.

isaacs:
I think this is a better API for custom tags.

PR #104, opened 2019-03-30
For Object, stringify mapping key as YAML rather than JSON

eemeli:
With this change, mapping keys are stringified during `YAML.parse()` or `YAML.Document#toJSON()` as YAML rather than JSON representations:

```js
YAML.parse('{ 1: 2 }: foo')
// was: { '{"1":2}': 'foo' }
// now: { '{ 1: 2 }': 'foo' }
```

The reason for this change is [this npm advisory](https://www.npmjs.com/advisories/788), which is related to this issue: nodeca/js-yaml#475 -- which also affects this library. Essentially, it's possible to abuse aliases to create an exponential tree of nodes consisting of collections that each have multiple references. When such a tree is used as the key of a mapping that is then converted to a JS object, `JSON.stringify()` was previously called on it, and the process runs out of memory creating the string.

Stringifying the key as YAML rather than JSON allows for the exponential explosion to be averted, as the repeated references use the same anchors as in the original input.

This change will be a part of the next minor release. It could arguably require a major version bump, but I figure that if you're working with structures that may include collections or aliases as keys and you care about the contents of said keys, you're almost certainly already using `mapAsMap: true` to avoid the stringification, and so this change should not break any reasonable implementations.

While this fix patches the library's own vulnerability to this type of attack, I'm considering whether I need to do something more when parsing this type of input, as the library user may not expect it to be able to return structures that are not safe to e.g. stringify as JSON.


eemeli:
Ended up adding a new option `maxAliasDepth`, and throwing an error during the document -> JS object stage when the reference depth exceeds the limit. By default three levels of aliasing are allowed, but that can be adjusted or disabled.

eemeli:
Ended up replacing the option with `maxAliasCount`, and actually tracking and limiting the count of how many times any one piece of data may be aliased rather than just the depth of the reference structure. By default, this is 100 (including the original), which ought to be enough for most cases, but still limiting enough that exponential or quadratic attacks are thwarted -- effectively the expansion is now at most linear at 100x the input size.

Setting `maxAliasCount` to 0 or 1 will now complain if the input contains any alias nodes.

PR #122, opened 2019-09-24
Add simpleKeys option

### LINKED ISSUE #88: null in object value
version 1.3.2 

```es6
const YAML = require('yaml');
console.log(YAML.stringify({all: {a: null, b: null, c: null }}));
```
it would get
```
all:
  ? a
  ? b
  ? c
```

why `?` is here? and it is not even a valid yaml.

I suppose it should be 

```yaml
all:
  a: 
  b: 
  c: 
```

my goal can be verify on
https://codebeautify.org/yaml-to-json-xml-csv
https://www.json2yaml.com/

eemeli:
Fixes #88 

Ping @seansfkelley; would this work for you? `simpleKeys` does a bit more than you asked for, as it'll not only prevent explicit keys like `? foo` from being used, but it'll also throw errors when a collection would be used as a key, e.g. `[ foo ]: bar`, which is otherwise valid YAML.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,7 @@ const defaultOptions = {
   mapAsMap: false,
   maxAliasCount: 100,
   prettyErrors: false, // TODO Set true in v2
+  simpleKeys: false,
   version: '1.2'
 }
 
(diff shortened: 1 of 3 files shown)


seansfkelley:
Yup, this should more than handle my use-case. I didn't even know some of those syntaxes were legal. o_O

Thanks for taking the time to add this option!

PR #137, opened 2020-02-16
[perf/stringify]: Use a Map for prevObjects in createNode

MikeRalphson:
Using the `Array.find` method for `prevObjects` seems to be a serious bottleneck in the `stringify` method (`src/schema/index` `createNode` method).

On a [14Mb YAML file](https://raw.githubusercontent.com/APIs-guru/openapi-directory/master/APIs/microsoft.com/graph/1.0.1/openapi.yaml), with this change, time to run `stringify` dropped from around 4 minutes to 4 seconds on my machine.

I get the same test results before and after.

### DIFF
--- a/src/schema/index.js
+++ b/src/schema/index.js
@@ -103,14 +103,14 @@ export default class Schema {
     }
     const obj = {}
     if (value && typeof value === 'object' && ctx.prevObjects) {
-      const prev = ctx.prevObjects.find(o => o.value === value)
+      const prev = ctx.prevObjects.get(value)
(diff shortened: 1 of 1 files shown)


eemeli:
Thank you, this is excellent! And apologies for taking so long to get to this, I was bogged down by other stuff and wanted to make sure that this would not cause any problems in browsers. But as it turns out, I'd forgotten that even IE 11 supports `Map`, so that's a non-issue.

I'll put together a new minor release shortly.

PR #145, opened 2020-03-21
Use named exports internally + update browser endpoints

eemeli:
It's tricky to get multi-platform targeting libraries right these days. Currently, `yaml` supports:

- CJS Node >=6
- ESM Node >=13
- Bundlers like Webpack & Rollup

To achieve that, the ESM sources are transpiled first to CJS under `dist/`, with a Node 6.5 target (actually giving us 6.0 support), and then to ESM under `browser/dist/`, with a Browserslist target that includes IE 11.

CJS Node finds its sources by just automatically adding a `.js` file extension, and then finding a manually crafted re-exporter from `dist/`. ESM Node goes via the `"exports"` object in the package.json, but does much the same. Both of these use the same CJS implementation internally.

Bundlers find their target via the `"browser"` package.json object, which leads to similar re-exporters from `browser/dist/` as for Node.

Eventually, I'd really like to be able to consolidate all of those into one ESM build, but that'll take some time yet for earlier Node versions to go out of favor, and possibly IE 11 as well -- and require a semver-major update to this library. Meanwhile, this PR is about taking a couple of small steps towards a bit more sanity.

Internally, this switches all exports and imports to be named, rather than defaults. This allows us to drop Webpack's `_interopRequire*` wrappers for `require()` calls, and not need an `__esModule` prop in each module.

The browser re-exporters are switched from CJS to ESM, as the implementation they're re-exporting is already ESM. This should be completely transparent to any normal users, but in the particular case of having a custom UMD build of just `yaml`, you may need to specify that you want its `default` export exposed. The Webpack option for this is `output.libraryExport: 'default'`.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,14 +1,15 @@
-const isBrowser = process.env.BABEL_ENV === 'browser'
-
-const envOptions = isBrowser
-  ? { modules: false, useBuiltIns: false }
-  : { targets: { node: '6.5' }, useBuiltIns: false }
-const presets = [['@babel/env', envOptions]]
-
 const plugins = [
   '@babel/plugin-proposal-class-properties',
   ['babel-plugin-trace', { strip: true }]
 ]
-if (isBrowser) plugins.push('@babel/plugin-transform-runtime')
+const envOptions = { modules: false, useBuiltIns: false }
 
+if (process.env.BABEL_ENV === 'browser') {
+  plugins.push('@babel/plugin-transform-runtime')
+} else {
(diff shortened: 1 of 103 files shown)


eemeli:
Comment on this by @BurtHarris [here](https://github.com/eemeli/yaml/commit/036d2044e8f5dfefba72603d3eafdc58bdae4133#commitcomment-38072455):
> @eemeli, I like this style, but could you explain what motivated the change?

I hope the PR's description answers your question?

Issue #53, opened 2018-11-28
Indentation of array elements is not standard


atombender:
Block sequences (ie., arrays) need not be indented. As the [1.1 spec says](http://yaml.org/spec/1.1/#id931893), the dash is *considered part of the indentation*. So this is the correct, idiomatic way of representing arrays:

``` yaml
array:
- one
- two
- three
```

The 1.2 spec does not explain this very well (it's not very well written as specs go), but this is the style used on the [current yaml.org front page](http://yaml.org/).

The current serialization adds one indentation level too many:

``` yaml
array:
  - one
  - two
  - three
```

This is admittedly a frequent point of confusion among YAML users; some people do indent. But as evidence of implementations that do it right, look at Ruby's standard `yaml` library. At least there should be a configuration option.

eemeli:
The most active public discussion on this can be found at prettier/prettier#4723, which indicated a clear preference for indentation. My gut feeling is that making this configurable would be more effort than it's worth -- YAML has plenty of corner cases where a change like this will trip you up.

However, if you're interested in putting together a PR I'd be willing to consider it. Probably the best place to start would be fiddling with the [`Collection`](../blob/master/src/schema/Collection.js) stringifier. Our test suite should be pretty good at catching most of the errors there.

atombender:
I wouldn't put much faith in random Internet polls by people I don't know, and Prettier has its share of [irrational decisions](https://github.com/prettier/prettier/issues/1985).

I'll see if I'll have time to make a PR. For now, I've switched to a different YAML library that does what I want.

eemeli:
I'm closing this, but would still be interested to later consider a PR for making the indenting behaviour optional. The right API for that would probably be via an `options` object added [here](../blob/master/src/schema/failsafe.js#L15).

PR #147, opened 2020-03-22
Add options to control indentation

### LINKED ISSUE #143: Is there an "Indent Children" Argument for Formatting
From what I understand, `prettier` uses this package to format `yaml` files but the default it formats by given children the same indentation as their parent.

```yaml
parent:
- child: 1
- child: 2
```

But I would like the format to indent all children:

```yaml
parent:
    - child: 1
    - child: 2
```

Is there a flag for this? Or is this specific to the the `prettier` project. 

For reference, I want to use this project for autoformatting but maintain the behavior of `yaml_formatter_indent_collection` flag in the [vim-yaml-formatter](https://github.com/tarekbecker/vim-yaml-formatter#configuration)

eemeli:
Closes #143
Related: #53

This adds the options `indent: number` and `indentSeq: boolean`, to allow for indentation to be customised. The defaults are `{ indent: 2, indentSeq: true }`.

`indent` is the number of spaces by which each indentation level is prefixed, and should be a positive integer.

If `indentSeq` is set to `false`, the `'- '` prefix of block sequence items is considered to be a part of the indentation where possible. This requires `indent >= 2`, and won't apply to sequences with anchors or explicit tags.

```js
const seq = YAML.createNode(['a'])
seq.commentBefore = 'sc'
const map = YAML.createNode({ foo: 'bar', seq })
map.commentBefore = 'mc'
const obj = { array: [{ a: 1, b: 2 }], map }

YAML.stringify(obj, { indent: 1 })
// array:
//  - a: 1
//    b: 2
// map:
//  #mc
//  foo: bar
//  seq:
//   #sc
//   - a

YAML.stringify(obj, { indent: 4, indentSeq: false })
// array:
//   - a: 1
//     b: 2
// map:
//     #mc
//     foo: bar
//     seq:
//         #sc
//       - a


eemeli:
Rebased to include updates to documentation.

Issue #47, opened 2018-11-20
Consider adding typings


julien-c:
either by exposing your public interface as a `.d.ts` file, or just renaming all your files to .ts =)

Thanks!

ikatyang:
See #23.

julien-c:
indeed, thanks

Issue #23, opened 2018-08-01
TypeScript Type declarations


ColinBradley:
TypeScript definition files or something in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) would be great. Do you have any plans for them?

I may create some definition files myself on DefinitelyTyped - as I would really like to use this package. Unless you have any objections? It would probably only be basic though.

Great package and great work on the documentation by the way! :)

ikatyang:
I've written some of them [here](https://github.com/ikatyang/yaml-unist-parser/blob/master/types/yaml/index.d.ts), I'm still waiting for stable release to finish the types and send to DT.

ColinBradley:
I've updated your implementation, @ikatyang. You can find it in this [gist](https://gist.github.com/ColinBradley/2109d049b6ff1ed00ea96e5944d92acd). There are still a few holes and it could do with being fully modularized into different files before going into DT, but it works for me! If I have some more time, I will try to get it into DT, but feel free to pick it up (or anyone else).

ikatyang:
Updated the updated implementation [here](https://github.com/ikatyang/yaml-unist-parser/blob/d928df79e5767b3db4d8fd6f8c58aa87f480fef8/types/yaml/index.d.ts) 😉, this should be all the types. I think sending PRs to DT before stable release is dangerous, there might be breaking changes which is not good for type consumers. (Actually, there's already one -> #25)

eemeli:
@ikatyang That does look almost complete. It's still missing the `yaml/scalar`, `yaml/types/binary` and `yaml/types/timestamp` exports, but that might be it. Plus of course the changes from at least #25, once I get a release together that includes that.

eemeli:
@ikatyang Now that the 1.0.0 release is out, the TS definitions could probably be released as well?

eemeli:
Closing this as the definitions have now been merged to the DT master.

PR #153, opened 2020-03-29
Add TypeScript typings

### LINKED ISSUE #102: TypeScript definitions are out of date
CC @ikatyang @ColinBradley

See also DefinitelyTyped/DefinitelyTyped#33780

The [DefinitelyTyped TS definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/yaml) haven't been updated since the 1.0 release last fall, and the library's API has expanded a bit since then. With the next minor release I'll be deprecating some of the endpoints and adding new ones thanks to #99.

Any suggestions for how to make it easier to keep these definitions up to date?

eemeli:
Fixes #102
Fixes DefinitelyTyped/DefinitelyTyped#33780
Closes DefinitelyTyped/DefinitelyTyped#43450
Related: #23, #47

These types are based on those published as `@types/yaml`, which are mostly the work of @ikatyang, with contributions from @ColinBradley and @shirk3y. In fact, I've picked out the original commits from DefinitelyTyped into this branch, and worked from there.

I'm relatively confident about these types, though I've not tested them very deeply. There are a few changes that may be relevant:
- The [DT types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yaml/index.d.ts) advertise themselves as being valid for TS 2.2; these require at the very least TS 2.3 due to `Iterable`, possibly a bit later than that.
- Some of the interface and namespaces are available from different names and endpoints compared to before:
    ```js
    // previously, with @types/yaml
    import { ast, cst, ParseOptions, YAMLError } from 'yaml'

    // now
    import { Options } from 'yaml'
    import { AST, YAMLError } from 'yaml/types'
    import { CST } from 'yaml/parse-cst'
    ```
- The TS class inheritance doesn't exactly match the JS implementation, but it's really close.
- I've not included typings for deprecated endpoints

The `tsconfig.json` and `yaml-tests.ts` files are remnants of the DT history, and should probably get removed before merging this.

Comments are very welcome here, in particular wrt changes to the DT types. For example, if there's reason to do so, re-exports could be added to match the previous typings.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,4 +1,6 @@
-import { cst, Document, Options, Tag } from './index'
+import { Document, Options, Tag } from './index'
+import { CST } from './parse-cst'
+import { Type } from './util'
 
 export interface BinaryOptions {
   /**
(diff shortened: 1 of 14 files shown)


types.d.ts
```diff
@@ -0,0 +1,392 @@
...
+     */
+    identify(value: any): boolean
+    /**
+     * The `Node` child class that implements this tag. Required for collections and tags that have overlapping JS representations.
+     */
+    nodeClass?: new () => any
+    /**
+     * Used by some tags to configure their stringification, where applicable.
+     */
+    options?: object
```
shirk3y: Should we define interface for these options?

eemeli: This needs to be extendable by custom tags, each of which will have their own options shape. I'm not sure what specifying this further would really help?

eemeli:
@ikatyang I added re-exports of the `AST` and `CST` namespaces to index.d.ts. All of the other types are tied pretty closely to the actual exports of `yaml/types` and `yaml/util`, so I don't think they really make sense being re-exported. Also good point about actually include the declarations in the package.

PR #161, opened 2020-04-26
Improve code & documentation around Pair.commentBefore

### LINKED ISSUE #157: Assigning commentBefore on a Pair
I want to add a pair to my document with a comment before the pair, e.g.:
```yaml
# This is a pair
foo: bar
```

I tried doing this using this code:
```ts
const pair = new Pair('foo', 'bar');
pair.commentBefore = ' This is a pair';
doc.add(pair);
```

But when I run the code, I get the following error:
```
(node:86824) UnhandledPromiseRejectionWarning: TypeError: Cannot create property 'commentBefore' on string 'foo'
    at Pair.set commentBefore [as commentBefore] (/path/to/my/project/node_modules/yaml/dist/schema/Pair.js:49:28)
```

As a more general question, how can I use YAML to create a document that has line comments preceding key/value pairs and/or line comments after a pair?  E.g., what's the best way to do something like this:
```yaml
# this is a comment about id before a pair
id: 123
# this is a comment about url before a pair
url: http://example.org/docs
# this is a comment after a pair that says you can optionally uncomment the next line
# debug: true

# this is a comment about author before a pair. I put this here to show that I
# can't use doc.comment to make that last comment (after a pair) work...
author: Joe Schmoe
```

using yaml 1.9.2

eemeli:
Fixes #157 

Following on from discussion with @cmoesel in the linked issue, this PR is doing two things:
1. Document the pre-existing `doc.schema.createPair(key, value)` method as an alternative way of creating a `Pair` object, which automatically wraps the `key` and `value` into nodes.
2. Be more explicit about `key` needing to be a node when getting or setting `pair.commentBefore`.

### DIFF
--- a/docs/05_content_nodes.md
+++ b/docs/05_content_nodes.md
@@ -194,7 +194,7 @@ doc.toString()
 
 To construct a `YAMLSeq` or `YAMLMap`, use [`YAML.createNode()`](#yaml-createnode) with array, object or iterable input, or create the collections directly by importing the classes from `yaml/types`.
 
-Once created, normal array operations may be used to modify the `items` array. New `Pair` objects may created by importing the class from `yaml/types` and using its `new Pair(key, value)` constructor.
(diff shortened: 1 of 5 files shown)


cmoesel:
Thank you, @eemeli!  Looks like a good solution!

PR #167, opened 2020-05-07
Add directivesEndMarker to Document type

Whoaa512:
Noticed this was missing when I tried to create a document in typescript on v1.9.2

Not sure if this is the right place to add or if the `.d.ts` files are generated in someway


eemeli:
Thank you! That's exactly the right fix for now. I wish that the d.ts files could be generated somehow, but the scripting for that turned out to be more difficult than doing the work manually.

PR #172, opened 2020-05-17
Resolve known tags in core schema

eemeli:
When using the `core` schema, if the parsed source includes one of the following explicit tags, parse it the same way as in the `yaml-1.1` schema, and include the relevant tag in `doc.schema.tags`:

- `!!binary`
- `!!omap`
- `!!pairs`
- `!!set`
- `!!timestamp`

To disable, use the option `{ resolveKnownTags: false }` in `YAML.parse*()`

#### Todo:
- [x] Update documentation
- [x] Review option fallback for schema-specific options

### DIFF
--- a/docs/03_options.md
+++ b/docs/03_options.md
@@ -20,25 +20,27 @@ YAML.Document.defaults
 
 The `version` option value (`'1.2'` by default) may be overridden by any document-specific `%YAML` directive.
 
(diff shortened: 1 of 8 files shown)


eemeli:
Documentation updated. Whether options should be refactored somehow is a rather separate question, so not going to block this PR any further on that.

PR #177, opened 2020-06-19
Improve quoted string flexibility

Himura2la:
We use this library to update files, which are also updated by another (YamlDotNet) library, and we have quoting style not in sync, which generates useless changes. I added an option to choose the default quoting style, and also tried to avoid escaped single quotes within string content.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -163,7 +163,7 @@ export namespace scalarOptions {
      */
     defaultType: Scalar.Type
     /**
-     * Use single quote (') as a default quoting style of strings
+     * Use 'single quote' rather than "double quote" by default
      *
      * Default: `false`
      */
--- a/index.d.ts
+++ b/index.d.ts
@@ -163,11 +163,11 @@ export namespace scalarOptions {
      */
     defaultType: Scalar.Type
     /**
-     * The default quoting style of strings
+     * Use single quote (') as a default quoting style of strings
      *
-     * Default: `'QUOTE_DOUBLE'`
+     * Default: `false`
      */
-    defaultQuote: Scalar.Type
+    defaultQuoteSingle: boolean
     doubleQuoted: {
       /**
        * Whether to restrict double-quoted strings to use JSON-compatible syntax.
--- a/index.d.ts
+++ b/index.d.ts
@@ -162,6 +162,12 @@ export namespace scalarOptions {
      * Default: `'PLAIN'`
      */
     defaultType: Scalar.Type
+    /**
+     * The default quoting style of strings
+     *
+     * Default: `'QUOTE_DOUBLE'`
+     */
+    defaultQuote: Scalar.Type
     doubleQuoted: {
       /**
        * Whether to restrict double-quoted strings to use JSON-compatible syntax.
--- a/src/stringify/stringifyString.js
+++ b/src/stringify/stringifyString.js
@@ -219,10 +219,12 @@ function plainString(item, ctx, onComment, onChompKeep) {
       value
     )
   ) {
+    const hasDouble = value.indexOf('"') !== -1
+    const hasSingle = value.indexOf("'") !== -1
     let quotedString
-    if (value.indexOf('"') !== -1 && value.indexOf("'") === -1) {
+    if (hasDouble && !hasSingle) {
       quotedString = singleQuotedString
-    } else if (value.indexOf("'") !== -1 && value.indexOf('"') === -1) {
+    } else if (hasSingle && !hasDouble) {
       quotedString = doubleQuotedString
     } else if (strOptions.defaultQuoteSingle) {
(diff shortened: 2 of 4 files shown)


Himura2la:
@eemeli, looks like the pipeline failed not because of my changes

src/stringify/stringifyString.js
```diff
@@ -204,7 +204,7 @@ function blockString({ comment, type, value }, ctx, onComment, onChompKeep) {
   return `${header}\n${indent}${body}`
 }
 
-function plainString(item, ctx, onComment, onChompKeep) {
+function plainString(item, ctx, onComment, onChompKeep, defaultQuote) {
```
eemeli, line 207: Why is `defaultQuote` being passed as an arg, rather than being read from `strOptions`?

src/stringify/stringifyString.js
```diff
@@ -219,16 +219,23 @@ function plainString(item, ctx, onComment, onChompKeep) {
...
     // - start with an indicator character (except [?:-]) or /[?-] /
     // - '\n ', ': ' or ' \n' anywhere
     // - '#' not preceded by a non-space char
     // - end with ' ' or ':'
     return implicitKey || inFlow || value.indexOf('\n') === -1
-      ? value.indexOf('"') !== -1 && value.indexOf("'") === -1
-        ? singleQuotedString(value, ctx)
-        : doubleQuotedString(value, ctx)
+      ? defaultQuotedString(value, ctx)
       : blockString(item, ctx, onComment, onChompKeep)
```
eemeli, line 239: With this added complexity, the ternaries should be dropped and an `if` statement used instead. Wanting to add a `prettier-ignore` comment is rather indicative of this.

This should help avoid useless computations by moving them to only the path where they're used, as well as not indexOf'ing value multiple times. Also, `defaultQuotedString` should probably rather be called `quotedString`.

src/tags/options.js
```diff
@@ -13,6 +13,7 @@ export const nullOptions = { nullStr: 'null' }
 
 export const strOptions = {
   defaultType: Type.PLAIN,
+  defaultQuote: Type.QUOTE_DOUBLE,
```
eemeli, line 16: Given that this only has two valid values, wouldn't it make sense to formulate it as a boolean instead?

Himura2la:
Thanks for your review, I'll change the code according to your comments ASAP

src/stringify/stringifyString.js
Himura2la: I did not notice that `strOptions` is available in this context. Will try to remove the parameter

src/stringify/stringifyString.js
Himura2la: OK, I expanded it into if/else chain, but I really not sure it became better

index.d.ts
```diff
@@ -162,6 +162,12 @@ export namespace scalarOptions {
      * Default: `'PLAIN'`
      */
     defaultType: Scalar.Type
+    /**
+     * Use single quote (') as a default quoting style of strings
```
eemeli, line 166:
```suggestion
     * Use 'single quote' rather than "double quote" by default
```

src/stringify/stringifyString.js
```diff
@@ -219,16 +219,24 @@ function plainString(item, ctx, onComment, onChompKeep) {
...
+    let quotedString
+    if (value.indexOf('"') !== -1 && value.indexOf("'") === -1) {
+      quotedString = singleQuotedString
+    } else if (value.indexOf("'") !== -1 && value.indexOf('"') === -1) {
+      quotedString = doubleQuotedString
+    } else if (strOptions.defaultQuoteSingle) {
+      quotedString = singleQuotedString
+    } else {
+      quotedString = doubleQuotedString
+    }
```
eemeli, line 231:
```suggestion
    const hasDouble = value.indexOf('"') !== -1
    const hasSingle = value.indexOf("'") !== -1
    let quotedString
    if (hasDouble && !hasSingle) {
      quotedString = singleQuotedString
    } else if (hasSingle && !hasDouble) {
      quotedString = doubleQuotedString
    } else if (strOptions.defaultQuoteSingle) {
      quotedString = singleQuotedString
    } else {
      quotedString = doubleQuotedString
    }
```

Himura2la:
Thank you for your suggestions, it's much better now

eemeli:
Thank you @Himura2la!

Himura2la:
@eemeli thanks for merging this helpful feature. When do you plan to make it available in npm?

eemeli:
Sorry, missed your last comment. I'll try and get a `2.0.0-alpha.1` release out within the next week or so.

Himura2la:
@eemeli, just a friendly bump: we can't close a ticket in our system unless this PR is released.

PR #186, opened 2020-08-05
Breaking changes to Document & createNode APIs

eemeli:
This PR combines a few closely related changes, and is intended to streamline how new documents and nodes are created.

### `new YAML.Document(value, options?)`

The Document constructor gets a new first argument `value`. This allows for much more ergonomic document creation:

```js
// before
const doc = new YAML.Document()
doc.setSchema()
doc.contents = doc.schema.createNode({ some: 'value' })

// after
const doc = new YAML.Document({ some: 'value' })
```
The `doc.schema` is now also left uninitialised only if `value === undefined`, and is automatically set if `createNode()` or `createPair()` are called.

### `doc.createNode(value, options?)`

The methods `YAML.createNode()` and `doc.schema.createNode()` are replaced by a single `doc.createNode()`. Their non-value arguments are also collected into an options object:

```js
// before
const doc = YAML.createDocument({ customTags: ['omap'] })
doc.setSchema()
const seq = YAML.createNode([1, 2, 3])
const omap = doc.schema.createNode([['foo', 13], ['bar', 42]], true, '!!omap')

// after
const doc = YAML.createDocument(null, { customTags: ['omap'] })
const seq = doc.createNode([1, 2, 3])
const omap = doc.createNode([['foo', 13], ['bar', 42]], { tag: '!!omap' })
```

Unlike its precedents, `doc.createNode()` is able to support circular references, which were previously only supported by `YAML.stringify()`.

### `doc.createPair(key, value, options?)`

The method `doc.schema.createPair()` is replaced by `doc.createPair()`, and also uses an options bag instead of named positional arguments.

```js
// before
const doc = YAML.createDocument()
doc.setSchema()
const pair = doc.schema.createPair('key', 'value')

// after
const doc = YAML.createDocument()
const pair = doc.createPair('key', 'value')
```

Internally, `doc.createPair()` is just sugar around two separate `doc.createNode()` calls and the `Pair` constructor.


eemeli:
Rebase to account for changes to master.

PR #188, opened 2020-08-27
Modify the output content in readme.

dota17:
Run the `YAML.stringify()` in readme and get the different result.


eemeli:
Thank you, good catch!

Issue #170, opened 2020-05-11
Remember source string for empty/null nodes


kasp1:
Hi there, I can't seem to be able to stringify `null` as expected, any idea what can be wrong?

repro:

```
const YAML = require('yaml')

let doc = YAML.parse('hi:')

console.log(doc) // 'Object {hi: null}'

// expected string: 'hi:'
console.log(YAML.stringify(doc)) // '? hi'

YAML.scalarOptions.null.nullStr = ''

console.log(YAML.stringify(doc)) // '? hi'

YAML.scalarOptions.null.nullStr = '~'

console.log(YAML.stringify(doc)) // '? hi'
```

eemeli:
The [option](https://eemeli.org/yaml/#options) to do what you want is `{ simpleKeys: true }`. This isn't the first time this is asked, so I think I need to find a way to improve the docs on this, to make it clearer/more discoverable.

eemeli:
Also, this is a bit besides what you're reporting, but the string representation of the `null` value isn't being retained at the AST level like it really ought to. Should fix that too.

kasp1:
Hi, thanks for your advice and for your library. :-)

eemeli:
Fixed in #193.

PR #193, opened 2020-09-04
Store the source string and outputs it in toString()/String().

dota17:
fix issue #170 .
+ Store the source string for null nodes.
+ Modify and supplement the test cases for `String()`.

### DIFF
--- a/src/resolve/resolveScalar.js
+++ b/src/resolve/resolveScalar.js
@@ -1,16 +1,10 @@
 import { Scalar } from '../ast/Scalar.js'
 
-// match custom string on null/bool
-function matchCustomTest(str, customTest) {
-  if (!customTest) return false
-  return customTest(str)
-}
-
 // falls back to string on no match
 export function resolveScalar(str, tags, scalarFallback) {
-  for (const { format, test, customTest, resolve } of tags) {
+  for (const { format, test, resolve } of tags) {
     if (test) {
-      const match = str.match(test) || matchCustomTest(str, customTest)
+      const match = str.match(test)
       if (match) {
         let res = resolve.apply(null, match)
         if (!(res instanceof Scalar)) res = new Scalar(res)
--- a/src/resolve/resolveScalar.js
+++ b/src/resolve/resolveScalar.js
@@ -1,10 +1,16 @@
 import { Scalar } from '../ast/Scalar.js'
 
+// match custom string on null/bool
+function matchCustomTest(str, customTest) {
+  if (!customTest) return false
+  return customTest(str)
+}
+
 // falls back to string on no match
 export function resolveScalar(str, tags, scalarFallback) {
-  for (const { format, test, resolve } of tags) {
+  for (const { format, test, customTest, resolve } of tags) {
     if (test) {
-      const match = str.match(test)
+      const match = str.match(test) || matchCustomTest(str, customTest)
       if (match) {
         let res = resolve.apply(null, match)
         if (!(res instanceof Scalar)) res = new Scalar(res)
--- a/src/tags/core.js
+++ b/src/tags/core.js
@@ -23,14 +23,10 @@ export const nullObj = {
     ctx.wrapScalars ? new Scalar(null) : null,
   default: true,
   tag: 'tag:yaml.org,2002:null',
-  sourceStr: null,
   test: /^(?:~|[Nn]ull|NULL)?$/,
-  customTest: str => {
-    return str === nullOptions.nullStr ? [str] : null
-  },
   resolve: str => {
     const node = new Scalar(null)
-    if(str) node.sourceStr = str
+    node.sourceStr = str
     return node
   },
   options: nullOptions,
--- a/src/tags/core.js
+++ b/src/tags/core.js
@@ -23,10 +23,18 @@ export const nullObj = {
     ctx.wrapScalars ? new Scalar(null) : null,
   default: true,
   tag: 'tag:yaml.org,2002:null',
+  sourceStr: null,
   test: /^(?:~|[Nn]ull|NULL)?$/,
-  resolve: () => null,
+  customTest: str => {
+    return str === nullOptions.nullStr ? [str] : null
+  },
+  resolve: str => {
+    const node = new Scalar(null)
+    if(str) node.sourceStr = str
+    return node
+  },
   options: nullOptions,
-  stringify: () => nullOptions.nullStr
+  stringify: ({ sourceStr }) => sourceStr ?? nullOptions.nullStr
 }
 
 export const boolObj = {
--- a/tests/doc/parse.js
+++ b/tests/doc/parse.js
@@ -75,6 +75,16 @@ describe('tags', () => {
   })
 })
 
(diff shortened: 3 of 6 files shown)


eemeli (changes requested):
This is great! The logic's not quite in the right place, but the idea is exactly right. I got stuck waaay deep when I started on this one myself, as there's some related improvements that could be made to the CST parser's handling of empty nodes.

I think we might want to do the same for booleans as well?

src/resolve/resolveScalar.js
```diff
@@ -2,13 +2,14 @@ import { Scalar } from '../ast/Scalar.js'
...
 export function resolveScalar(str, tags, scalarFallback) {
-  for (const { format, test, resolve } of tags) {
+  for (const { format, sourceStr, test, resolve } of tags) {
     if (test) {
       const match = str.match(test)
       if (match) {
         let res = resolve.apply(null, match)
         if (!(res instanceof Scalar)) res = new Scalar(res)
         if (format) res.format = format
+        if (typeof sourceStr !== 'undefined') res.sourceStr = str
```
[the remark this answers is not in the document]
dota17: Nice code!! Done.

src/tags/core.js
```diff
@@ -23,10 +23,11 @@ export const nullObj = {
     ctx.wrapScalars ? new Scalar(null) : null,
   default: true,
   tag: 'tag:yaml.org,2002:null',
+  sourceStr: null,
   test: /^(?:~|[Nn]ull|NULL)?$/,
   resolve: () => null,
   options: nullOptions,
-  stringify: () => nullOptions.nullStr
+  stringify: ({ sourceStr }) => sourceStr ?? nullOptions.nullStr
```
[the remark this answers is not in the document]
dota17: + You mean "sourceStr" is not quite suitable? How about "originalStr".
+ And, I have add the codes and testcase to reparse the string which matches the defined custom null string, 
+ But it can not handle the `"!!null str"` or`"a: \"None\""` when
> YAML.scalarOptions.null.nullStr = 'None'

I think it is normal.

dota17:
> I think we might want to do the same for booleans as well?

Yes, I've thought about `bool`. It will be finished later.

src/resolve/resolveScalar.js
```diff
@@ -1,10 +1,16 @@
...
+  return customTest(str)
+}
+
 // falls back to string on no match
 export function resolveScalar(str, tags, scalarFallback) {
-  for (const { format, test, resolve } of tags) {
+  for (const { format, test, customTest, resolve } of tags) {
     if (test) {
-      const match = str.match(test)
+      const match = str.match(test) || matchCustomTest(str, customTest)
```
eemeli, line 13: This still isn't the right place for these checks. Really there's no need to modify `resolveScalar` at all here.

src/tags/core.js
```diff
@@ -23,10 +23,18 @@ export const nullObj = {
     ctx.wrapScalars ? new Scalar(null) : null,
   default: true,
   tag: 'tag:yaml.org,2002:null',
+  sourceStr: null,
   test: /^(?:~|[Nn]ull|NULL)?$/,
-  resolve: () => null,
+  customTest: str => {
+    return str === nullOptions.nullStr ? [str] : null
+  },
```
eemeli, line 30: This `customTest` approach makes it far too easy to produce invalid YAML just by using a bad `nullStr` option. It might be fine to _output_  the bad YAML if that's what the user really wants, but that doesn't mean that we should just accept it as input. To handle that, they should modify the `test` regexp that their schema uses.

src/tags/core.js
```diff
@@ -23,10 +23,18 @@ export const nullObj = {
     ctx.wrapScalars ? new Scalar(null) : null,
   default: true,
   tag: 'tag:yaml.org,2002:null',
+  sourceStr: null,
```
eemeli, line 26: I don't think this has any role anymore?

src/tags/core.js
```diff
@@ -23,10 +23,18 @@ export const nullObj = {
...
   test: /^(?:~|[Nn]ull|NULL)?$/,
-  resolve: () => null,
+  customTest: str => {
+    return str === nullOptions.nullStr ? [str] : null
+  },
+  resolve: str => {
+    const node = new Scalar(null)
+    if(str) node.sourceStr = str
+    return node
+  },
```
eemeli, line 35: Is there a particular reason why the empty string `''` could not be used as the `sourceStr` value? It's a valid YAML representation of `null`.

tests/doc/parse.js
```diff
@@ -75,6 +75,16 @@ describe('tags', () => {
...
 
+describe('custom string on node', () => {
+  test('custom null', () => {
+    YAML.scalarOptions.null.nullStr = 'None'
+    const doc = YAML.parse('a: null')
+    const str = YAML.stringify(doc, {simpleKeys: true})
+    expect(str).toBe('a: None\n')
+    expect(YAML.parse(str)).toEqual({a: null})
+  })
+})
```
eemeli, line 86: I don't think we should include an example that uses invalid YAML. Use for example `~` instead of `None`.

eemeli (changes requested):
Getting there...

src/tags/core.js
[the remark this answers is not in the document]
eemeli: Let's go with `sourceStr` for now.

There's no need to easily support a non-YAML representation of `null` like `'None'`.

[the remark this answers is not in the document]
dota17: > it should be tested with the tag's regexp to make sure that it'll re-parse as null

oops, I misunderstood it.

src/resolve/resolveScalar.js
dota17: Removed.

src/tags/core.js
dota17: Removed.

src/tags/core.js
dota17: ok, done

src/tags/core.js
dota17: yes, removed.

tests/doc/parse.js
dota17: ok, add the testcase for `~` and `''`

dota17:
@eemeli 
Are my changes acceptable? If so, I will submit the changes for bool.

eemeli (approved):

eemeli:
Looks good! I'm now wondering though if the right generalisation would be to apply this to all scalars? As in, always store the `sourceStr`, and when stringifying, check if it'll still resolve to the same value. That might need to have an option to toggle it though, if it's as bad a performance hit as I suspect it might be.

OTOH, just applying this to nulls and booleans would cover a significant fraction of the real-world use cases.

@dota17, for the boolean solution we'll need to add a re-parsing check, as otherwise converting from YAML 1.1 to YAML 1.2 would not convert `no` to `false` like it should. For nulls triggering that error would require you to e.g. manually set `sourceStr` to something invalid, but we should probably cover that as well if we're doing it for booleans.

PR #195, opened 2020-09-14
Dump the simple key which is over 1024 properly

dota17:
the single line scalar which is simple key  not span more than 1024 Unicode characters on `String()`/`stringify()`

### DIFF
--- a/src/ast/Pair.js
+++ b/src/ast/Pair.js
@@ -96,6 +96,10 @@ export class Pair extends Node {
         const msg = 'With simple keys, collection cannot be used as a key value'
         throw new Error(msg)
       }
+      if (key instanceof Scalar && key.value?.length > 1024) {
+        const msg = 'With simple keys, single line scalar must not span more than 1024 characters'
+        throw new Error(msg)
+      }
     }
     const explicitKey =
       !simpleKeys &&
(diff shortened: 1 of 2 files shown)


src/ast/Pair.js
```diff
@@ -96,14 +96,19 @@ export class Pair extends Node {
...
     }
     const explicitKey =
       !simpleKeys &&
       (!key ||
         keyComment ||
         key instanceof Collection ||
         key.type === Type.BLOCK_FOLDED ||
-        key.type === Type.BLOCK_LITERAL)
+        key.type === Type.BLOCK_LITERAL ||
+        (key instanceof Scalar && key.value?.length > 1024))
```
eemeli, line 111: We can't trust `key.value.length` to match the actual character count that will get printed. Same goes for the `simpleKeys` case above.

For instance, `key.value` can be a BigNum with a string representation of more than 1024 digits, but it does not have a `length` property. Neither can we trust `String(key.value)` to be an accurate representation, as the schema may include a custom tag that'll use some other way of representing the value.

Because anything that can be represented as an implicit key can also be represented as an explicit key, the right place to control this is around here: https://github.com/eemeli/yaml/blob/cc4456c254bc2476cd837b8418c3349f4576e890/src/ast/Pair.js#L127

At that point, if `explicitKey` is `false` and `str.length > 1024`, we can either complain if `simpleKeys` is set, or use the explicit form.

eemeli (changes requested):
Good catch!

src/ast/Pair.js
dota17: Nice, thoughtful idea!! How about the new codes?

eemeli (approved):

eemeli:
Looks good!

PR #199, opened 2020-09-25
Support to parse the string which matches the defined custom bool string

dota17:
* Store the source string for boolnodes.
* Modify and supplement the test cases for String().

### DIFF
--- a/src/tags/core.js
+++ b/src/tags/core.js
@@ -17,6 +17,15 @@ function intStringify(node, radix, prefix) {
   return stringifyNumber(node)
 }
 
+function stringifyBool(node) {
+  const { value, sourceStr } = node
+  if (sourceStr) {
+    const match = boolObj.test.test(sourceStr)
+    if (match && value === (sourceStr[0] === 't' || sourceStr[0] === 'T')) return sourceStr
+  }
+  return value ? boolOptions.trueStr : boolOptions.falseStr
+}
+
 export const nullObj = {
   identify: value => value == null,
   createNode: (schema, value, ctx) =>
@@ -38,9 +47,13 @@ export const boolObj = {
   default: true,
   tag: 'tag:yaml.org,2002:bool',
   test: /^(?:[Tt]rue|TRUE|[Ff]alse|FALSE)$/,
-  resolve: str => str[0] === 't' || str[0] === 'T',
+  resolve: str => {
+    const node = new Scalar(str[0] === 't' || str[0] === 'T')
+    node.sourceStr = str
+    return node
+  },
   options: boolOptions,
-  stringify: ({ value }) => (value ? boolOptions.trueStr : boolOptions.falseStr)
+  stringify: stringifyBool
 }
 
 export const octObj = {
--- a/src/tags/yaml-1.1/index.js
+++ b/src/tags/yaml-1.1/index.js
@@ -10,8 +10,37 @@ import { pairs } from './pairs.js'
 import { set } from './set.js'
 import { intTime, floatTime, timestamp } from './timestamp.js'
 
-const boolStringify = ({ value }) =>
-  value ? boolOptions.trueStr : boolOptions.falseStr
+const boolStringify = ({ value, sourceStr }) => {
+  const boolObj = value ? trueObj : falseObj
+  if (sourceStr && boolObj.test.test(sourceStr)) return sourceStr
+  return value ? boolOptions.trueStr : boolOptions.falseStr
+}
+
+const boolResolve = (value, str) => {
+  const node = new Scalar(value)
+  node.sourceStr = str
+  return node
+}
+
+const trueObj = {
+  identify: value => value === true,
+  default: true,
+  tag: 'tag:yaml.org,2002:bool',
+  test: /^(?:Y|y|[Yy]es|YES|[Tt]rue|TRUE|[Oo]n|ON)$/,
+  resolve: str => boolResolve(true, str),
+  options: boolOptions,
+  stringify: boolStringify
+}
+
+const falseObj = {
+  identify: value => value === false,
+  default: true,
+  tag: 'tag:yaml.org,2002:bool',
+  test: /^(?:N|n|[Nn]o|NO|[Ff]alse|FALSE|[Oo]ff|OFF)$/i,
+  resolve: str => boolResolve(false, str),
+  options: boolOptions,
+  stringify: boolStringify
+}
 
 const intIdentify = value =>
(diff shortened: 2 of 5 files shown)


src/tags/core.js
```diff
@@ -17,6 +17,16 @@ function intStringify(node, radix, prefix) {
   return stringifyNumber(node)
 }
 
+function stringifyBool(node) {
+  const { value } = node
+  let { sourceStr } = node
+  if (sourceStr) {
+    const match = sourceStr.match(/^(?:[Tt]rue|TRUE|[Ff]alse|FALSE)$/)
```
eemeli, line 24: 1. There should not be a reason to repeat the RegExp here; it's already available as `nullObj.test`, and that way there's no danger of this becoming unconnected if the user modifies the tag.
2. If you're not using the result, use `re.test(str)` instead.

src/tags/core.js
```diff
@@ -17,6 +17,16 @@ function intStringify(node, radix, prefix) {
...
+    if (!match) sourceStr = null
+  }
+  return value ? sourceStr ?? boolOptions.trueStr : sourceStr ?? boolOptions.falseStr
```
eemeli, line 27: I actually like ternaries, but five `?` on a single line is perhaps too much even for me.

I'm also not at all sure that this is right. What happens if a node is parsed with a `true` value, updated to `false` by the user, and the doc is then re-stringified? There should be a test case for that.

src/tags/yaml-1.1/index.js
```diff
@@ -10,8 +10,14 @@ import { pairs } from './pairs.js'
 import { set } from './set.js'
 import { intTime, floatTime, timestamp } from './timestamp.js'
 
-const boolStringify = ({ value }) =>
-  value ? boolOptions.trueStr : boolOptions.falseStr
+const boolStringify = ({ value, sourceStr }) =>
```
eemeli, line 13: Why is the `yaml-1.1` behaviour different from `core`?

tests/doc/parse.js
```diff
@@ -91,6 +91,24 @@ describe('custom string on node', () => {
...
+  })
+
+  test('empty string upperBool', () => {
+    YAML.scalarOptions.bool.trueStr = 'TRUE'
+    YAML.scalarOptions.bool.falseStr = 'FALSE'
+    const doc = YAML.parse('[true, false]')
+    const str = YAML.stringify(doc)
+    expect(str).toBe('- TRUE\n- FALSE\n')
+    expect(YAML.parse(str)).toEqual([true, false])
+  })
```
eemeli, line 111: What's the new ground being covered by these tests?

eemeli (changes requested):

src/tags/core.js
dota17: done

src/tags/core.js
dota17: done

tests/doc/parse.js
dota17: This test is to check whether the program can output correctly after the user sets the scalarOptions.

src/tags/yaml-1.1/index.js
dota17: Because we don't need to check whether the sourceStr matches the bool format in Yaml1.1 schema. THe bool format which Yaml1.1 schema supports is more than core schema.

src/tags/core.js
eemeli: I'm still seeing five `?` on a single line. Rather than setting `sourceStr = null` on line 25, invert the test and do an early `return sourceStr` from there.

tests/doc/parse.js
eemeli: Okay, so they should then go with the stringify tests, rather than here.

src/tags/yaml-1.1/index.js
```diff
@@ -10,8 +10,18 @@ import { pairs } from './pairs.js'
 import { set } from './set.js'
 import { intTime, floatTime, timestamp } from './timestamp.js'
 
-const boolStringify = ({ value }) =>
-  value ? boolOptions.trueStr : boolOptions.falseStr
+const boolStringify = ({ value, sourceStr }) => {
+  const match = yaml11.filter(t => t.identify && t.identify(value))
+  const tagObj = match.find(t => !t.format)
+  if (sourceStr && !tagObj.test.test(sourceStr)) sourceStr = null
+  return value ? sourceStr ?? boolOptions.trueStr : sourceStr ?? boolOptions.falseStr
```
eemeli, line 17: If `sourceStr` is not set, this does a whole lot of useless extra work. Also, if there's a need to easily get at the right regexp, just extract them from the tag objects into top-level variables and use them from there.

eemeli (changes requested):
Resolution is starting to look good, but the stringifiers still need some tweaking.

src/tags/core.js
dota17: Good! done.

tests/doc/parse.js
dota17: Good! done.

src/tags/yaml-1.1/index.js
```diff
@@ -64,20 +74,20 @@ export const yaml11 = failsafe.concat(
       stringify: ({ sourceStr }) => sourceStr ?? nullOptions.nullStr
     },
     {
-      identify: value => typeof value === 'boolean',
+      identify: value => typeof value === 'boolean' && value,
```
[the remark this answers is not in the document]
dota17: Good! done.

src/tags/yaml-1.1/index.js
```diff
@@ -64,20 +74,20 @@ export const yaml11 = failsafe.concat(
...
       tag: 'tag:yaml.org,2002:bool',
       test: /^(?:Y|y|[Yy]es|YES|[Tt]rue|TRUE|[Oo]n|ON)$/,
-      resolve: () => true,
+      resolve: str => boolResolve(true, str),
       options: boolOptions,
       stringify: boolStringify
     },
     {
-      identify: value => typeof value === 'boolean',
+      identify: value => typeof value === 'boolean' && !value,
```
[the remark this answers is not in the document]
dota17: Good! done.

src/tags/yaml-1.1/index.js
dota17: Good! done.

dota17:
@eemeli
I've modified the code as you asked. Do you have any suggestions for the new changes？

eemeli (approved):
Sorry for taking so long to get at this, and thank you for reminding me about it!

PR #203, opened 2020-10-05
Refactor parsing completely

### LINKED ISSUE #116: Preserve newlines within comments
Would it be possible to maintain newlines within comments? This would allow the structure of a lot of documentation to be adhered to more closely.

eemeli:
This PR will effectively replace everything in `/src/cst/` and `/src/resolve/` with new code, and redefine the CST level API. As such, I'm guessing that at least @ikatyang and @iann0036 might be interested in getting an early heads-up. This is a work in progress, and not complete yet. I'll update this PR & description as the work continues over the next few weeks (not promising anything about a timetable here).

There are three main reasons and starting points for this change.
1. Supporting Node.js stream input. I've previously tried to hack at this on and off over the past year or so, but didn't really get a breakthrough until about two weeks ago.
2. Untangling & re-architecting things to make them simpler to understand and maintain.
3. Preparing for future YAML spec updates. In particular, yaml/yaml-spec#59 is looking to apply [this change](https://github.com/yaml/yaml-spec/blob/RFC-directives-are-stream/rfc/RFC-directives-are-stream.md).

To that end, `/src/parse/parser` is a new parser that generates CST node trees from its input, internally using `/src/parse/lexer` to split the input into tokens. They're both capable of dealing with input coming in chunks, and `/src/parse/parse-stream` is a possible Node.js stream wrapper for the parser. See the parser source for TypeScript interfaces for the tokens.

Internally, "parsing" now means producing the CST from string input, while the Document object creation is called "composing". That should more closely match the language used in the YAML spec & other implementations. Currently, this composition starts from `/src/compose/compose-doc`, and `/src/compose/parse-docs` is a possible wrapper for connecting the parser outputs to the composition.

Thanks to all these changes, this finally fixes #116 and makes sure that the `null` nodes mentioned in #170 are removed from both the CST and AST levels.

As a point of interest, it should be pretty easy to use the lexer in order to build a really fast basic YAML highlighter. If you want to do fancier things like color-coding depending on the level of logical indentation, you'll probably need the parser as well, and for errors the composer too.

If you'd like to try this out, check out the `stream` branch and run these commands in the repo root:
```
npm install
npm run build:dev
npm start
```

### TODO

- [x] Parse string input into a CST
- [x] Compose scalars
- [x] Compose block collections
- [x] Compose flow collections
- [x] Resolve non-local tags (may require a `YAML.Document` API change)
- [x] Require map keys to be on a single line
- [x] Require map keys to have max 1024 characters
- [x] Assign props correctly between a map & its first key
- [ ] ~~Fix self-referential aliases on custom collections~~ (see #229)
- [x] Complain about unindented mapping content
- [ ] Add a public Node.js transform stream
- [x] Drop old code
- [x] Refactor the build to support the new JS/TS mix
- [ ] Update documentation


fkirc:
Hi,

I am interested in a specific use case:
I need to modify existing yml-files at several places, but I want to do it in a surgical way, with as little changes as possible.
This means that I want to:
- Preserve (almost) all comments
- Preserve (almost) all new lines, indentations and other whitespaces
- Preserve general structure and ordering as much as possible

Is the CST-API something that you would recommend for this use case?
If yes, does this PR affect this use case?

fkirc:
Never mind, I wrote a separate issue about whitespace preservation: https://github.com/eemeli/yaml/issues/207

I needed to keep CST-nodes, although I did not really touch the CST-API.

eemeli:
Got stuck/sidetracked around flow collection composition for _ages_, but that looks like it's starting to work now.

eemeli:
Some progress. The lexer had some corner cases that lead to an infinite loop; fixed those. Now, running the full test suite at least completes, but the new code is still failing 306/1962 tests.

iann0036:
Tried the branch, working well for my simple tests with parse/stringify and document parsing. Are there build instructions for the CST Parser (or similar replacement)?

eemeli:
@iann0036 I'm not certain yet what the public API for the CST parser will look like, but internally it's available from `src/parse/parser.ts`, and you can look at `src/compose/parse-docs.ts` for the scaffolding that I've now got around it to replicate the functionality of `YAML.parseAllDocuments()`.

If you're testing from outside the repo itself, you should e.g. modify the `"./parse-cst"` entry in the package.json `"exports"` config to point at `./lib/parse/parser.js`, so you can then:
```js
import { Parser } from 'yaml/parse-cst'
```

Apologies for the bad ergonomics, this is still a work in progress. Now down to 131 failing tests.

iann0036:
Cheers @eemeli! That's perfect.

Do you have, or intend to make, a direct stringify function for a CST/Parser Document? (i.e. one that preserves whitespace etc.). If not, happy to create one myself (and perhaps PR in), just didn't want to double up the work.

iann0036:
Initial work without diving too far into edges:

```
_addSourceToStr(item, property) {
    if (!item[property]) {
        return "";
    }

    if (!Array.isArray(item[property])) {
        item[property] = [ item[property] ];
    }

    var ret = "";

    for (var el of item[property]) {
        ret += el.source;
    }
    
    return ret;
}

stringifyItem(item) {
    var ret = "";

    ret += this._addSourceToStr(item, 'start');
    ret += this._addSourceToStr(item, 'key');
    ret += this._addSourceToStr(item, 'sep');
    if (item.value) {
        if (!Array.isArray(item.value)) {
            item.value = [ item.value ];
        }
        for (var value of item.value) {
            ret += this.stringifyItem(value);
        }
    }
    if (item.source) {
        if (item.props) {
            for (var prop of item.props) {
                ret += prop.source;
            }
        }
        ret += item.source;
    }
    if (item.items) {
        for (var subitem of item.items) {
            ret += this.stringifyItem(subitem);
        }
    }
    ret += this._addSourceToStr(item, 'end');

    return ret;
}
```

eemeli:
I hadn't actually considered a CST stringifier yet, but one will definitely need to be provided. The node structure is pretty stable, but it's possible that it'll still need to be tweaked, so I at least likely won't start on it until later.

Happy to consider a PR for it, but only after this massive PR is merged first. Given that the rest of the CST parser is now in TS, the stringifier will almost certainly need to be as well.

eemeli:
27 failing tests remain, all regarding comments & whitespace.

eemeli:
Finally about done with this. The ✅ is hiding the fact that some of the tests are skipped, but those are really only catching the lack of prettified errors + some remaining differences in how comments are assigned to nodes.

I may end up merging this and doing the remaining work in separate PRs.

eemeli:
I'm going to merge this, and update the docs later. I'm also not quite sure yet of the stream API, so that's getting left out for now. My current draft of that is available here: https://gist.github.com/eemeli/2436bda974549eb24ee0696bd8f6368b

The new CST level API is included, though, as the `Lexer` and CST `Parser` are now publicly exported, along with the AST/Document `Composer`. For now, see the documentation comments in the source for more details on their usage.

Issue #207, opened 2020-10-28
Allow source modifications with minimal changes


fkirc:
First of all, thank you for writing this yaml package, it works great 👍
The yaml package has become a critical part of  `attranslate`: https://github.com/fkirc/attranslate

`attranslate` has the requirement of "whitespace and comment preservation".
This means that `attranslate` compares two different YML-trees, modifies values of scalar nodes, and then serialize a file again with as little changes as possible.

I was able to implement this behavior with the following recursive tree-walk: https://github.com/fkirc/attranslate/blob/master/src/file-formats/yaml/yaml-manipulation.ts

However, there are still some problems.
The preservation of comments and single-linebreaks works correctly, but the preservation of multi-linebreaks doesn't work.

Could you point me in a direction to preserve multi-linebreaks?

eemeli:
Supporting that at the document level would require the data on multiple linebreaks to be retained somewhere, which it currently isn't. You may also find that preserving variable indent levels to require data that isn't currently retained.

The API tries to balance things between power and simplicity, and that does mean that some compromises are made; one such compromise is the expressibility of line breaks preceding a node, particularly in association with comment lines. If you would like for this API to change and thereby be able to retain additional information, I think you ought to propose what the alternative API ought to look like, and how it would work?

fkirc:
I understand that such features should not lead to an excessive complication of parser-internals, therefore I will try to keep it simple.

Right now, I am doing the following to preserve comments and single-linebreaks (this code-snippet is taken from https://github.com/fkirc/attranslate/blob/master/src/file-formats/yaml/yaml-manipulation.ts):

```TypeScript
  if (context.oldTargetNode?.comment) {
    node.comment = context.oldTargetNode.comment;
  }
  if (context.oldTargetNode?.commentBefore) {
    node.commentBefore = context.oldTargetNode.commentBefore;
  }
  if (context.oldTargetNode?.spaceBefore) {
    node.spaceBefore = context.oldTargetNode.spaceBefore;
  }
```

As we can see, this is a selective transfer of properties from one Node to another Node (during a recursive double-tree-walk).

Because of this working snippet, I am thinking about augmenting `Node` with additional properties:

- Preceding whitespace lines (could be a string-array)
- Variable indent levels, as long as it doesn't break YML-syntax

fkirc:
One more sidenote:

I also experimented with transferring CST-nodes from one Node to another Node:

```TypeScript
  if (context.oldTargetNode?.cstNode) {
[...]

PR #209, opened 2020-11-07
Add support for multi-linebreak-preservation

fkirc:
This is a solution for https://github.com/eemeli/yaml/issues/207.
I already tested it successfully on the following feature-branch of `attranslate`: https://github.com/fkirc/attranslate/pull/64
This API-proposal isn't really clean, but I do not dare to make any breaking changes as an external contributor.

### DIFF
--- a/docs/05_content_nodes.md
+++ b/docs/05_content_nodes.md
@@ -13,6 +13,8 @@ class Node {
       // into this node (undefined for pairs or if not parsed)
   spaceBefore: ?boolean,
       // a blank line before this node and its commentBefore
+  spacesBefore: ?number,
+      // the number of blank lines before this node and its commentBefore
   tag: ?string,       // a fully qualified tag, if required
   toJSON(): any       // a plain JS or JSON representation of this node
 }
@@ -243,6 +245,7 @@ const doc = YAML.parseDocument('[ one, two, three ]')
 
 doc.contents.items[0].comment = ' item comment'
 doc.contents.items[1].spaceBefore = true
+doc.contents.items[1].spacesBefore = 1
 doc.comment = ' document end comment'
 
 doc.toString()
--- a/src/ast/Collection.js
+++ b/src/ast/Collection.js
@@ -146,8 +146,13 @@ export class Collection extends Node {
     const nodes = this.items.reduce((nodes, item, i) => {
       let comment
       if (item) {
-        if (!chompKeep && item.spaceBefore)
+        if (!chompKeep && item.spacesBefore) {
+          for (let count = 0; count < item.spacesBefore; count++) {
+            nodes.push({ type: 'comment', str: '' })
+          }
+        } else if (!chompKeep && item.spaceBefore) {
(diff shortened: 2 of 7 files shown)


eemeli:
I don't think I like this. On a conceptual level, the AST does not promise to retain the exact shape of the source string. Once it's been passed through once, it'll stay pretty close but even then the docs do include at least [this note](https://eemeli.org/yaml/#comments):
> Due to implementation details, the library's comment handling is not completely stable. In particular, when creating, writing, and then reading a YAML file, comments may sometimes be associated with a different node.

So in a sense the library is usable as a YAML prettifier by parsing & stringifying a source string. And as a part of that, it currently cleans up extraneous white space between nodes by collapsing it into a single empty line. This PR muddies the situation by changing the logic to only dropping any trailing white space from those lines, while retaining the complete drop of any empty lines after a comment.

I think I'd like to see much wider support for this change before considering it further. Also, there's no realistic way of making it not a breaking change, and having both `spaceBefore` and `spacesBefore` is just not going to happen.

fkirc:
> I don't think I like this. On a conceptual level, the AST does not promise to retain the exact shape of the source string. Once it's been passed through once, it'll stay pretty close but even then the docs do include at least [this note](https://eemeli.org/yaml/#comments):
> 
> > Due to implementation details, the library's comment handling is not completely stable. In particular, when creating, writing, and then reading a YAML file, comments may sometimes be associated with a different node.
> So in a sense the library is usable as a YAML prettifier by parsing & stringifying a source string. And as a part of that, it currently cleans up extraneous white space between nodes by collapsing it into a single empty line. This PR muddies the situation by changing the logic to only dropping any trailing white space from those lines, while retaining the complete drop of any empty lines after a comment.

I understand that this is not the way that things are currently working, therefore I think that we need to step back from this specific PR and instead debate on the general goals of this library.
Preserving the visual presentation of a YAML as much as possible is a valid use case, whereas "prettifying" is another valid use case.
Some users want to have a "prettifier" that aggressively reformats YAML to match some "standard form".
This is not the thing that `attranslate` wants, `attranslate` wants the former use case instead (preserve as much structure as possible).
So the crucial question is:

Do you want to support the "preservation use case", or do you only care about the "prettier use case"?

> I think I'd like to see much wider support for this change before considering it further. Also, there's no realistic way of making it not a breaking change, and having both `spaceBefore` and `spacesBefore` is just not going to happen.

To make it not a breaking change, we would probably need an additional config-option.
I would add such config-option if there is a chance of getting something merged (not necessarily this API-proposal).

fkirc:
Perhaps I should rephrase my question:

Is there anything I could do to improve support for the "preservation use case", without threatening the already existing "prettier use case"?

My point is not to merge this particular PR, but rather to find a direction that leads to an improved format-preservation.

fkirc:
I am closing this PR because it is not a nice fit for the current API.
Nevertheless, I am still aiming to get this feature done in another form.

PR #210, opened 2020-11-18
Add missing extension to Document.js import path

SyrupThinker:
A quick grep yields this as the only import with a missing extension.
This change allows a successful import with Deno.


eemeli (approved):

eemeli:
Thank you! Deno-compatibility is indeed intentional; this was an oversight.

Before making the support official, I'm still waiting for proper answers to the questions I posed here: https://github.com/denoland/deno/issues/3196#issuecomment-633142656.

Issue #208, opened 2020-10-30
Failed to use yaml lib with webpack 5


igorlysak:
Got an error during `webpack server` command. And I can't figure out what I do wrong. If possible could you please help?
```
> webpack serve

ℹ ｢wds｣: Project is running at http://localhost:3333/
ℹ ｢wds｣: webpack output is served from undefined
ℹ ｢wds｣: Content not from webpack is served from /Users/igor/Projects/yaml-test/src
✖ ｢wdm｣: 1 asset
46 modules

ERROR in ./node_modules/yaml/browser/dist/index.js 1:0
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> import { d as defaultTagPrefix, _ as _createForOfIteratorHelper, a as _typeof, b as _createClass, c as _classCallCheck, e as _defineProperty, Y as YAMLSyntaxError, T as Type, f as YAMLWarning, g as YAMLSemanticError, h as _slicedToArray, i as YAMLError, j as _inherits, k as _createSuper } from './PlainValue-ff5147c6.js';
| import { parse as parse$1 } from './parse-cst.js';
| import { b as binaryOptions, a as boolOptions, i as intOptions, n as nullOptions, s as strOptions, N as Node, P as Pair, S as Scalar, c as stringifyString, A as Alias, Y as YAMLSeq, d as YAMLMap, M as Merge, C as Collection, r as resolveNode, e as isEmptyPath, t as toJSON, f as addComment } from './resolveSeq-04825f30.js';

webpack 5.3.2 compiled with 1 error in 4332 ms
ℹ ｢wdm｣: Failed to compile.
```

Webpack config:
```js
const HtmlWebpackPlugin = require('html-webpack-plugin')

const path = require('path');

const srcDir = path.resolve(__dirname, 'src');

module.exports = {
  entry: `${srcDir}/index.js`,
  output: {
    path: path.resolve(__dirname, 'dist'),
    filename: '[name].[contenthash].js',
  },
  devtool: 'source-map',
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader',
          options: {
            presets: ['@babel/preset-env']
          }
        }
      }
    ]
  },
  plugins: [
    new HtmlWebpackPlugin({template: 'src/index.html'}),
  ],
[...]

PR #223, opened 2021-01-29
Drop 'yaml/parse-cst' endpoint

eemeli:
### BREAKING CHANGE
Users will need to update their code:

```diff
-import parseCST from 'yaml/parse-cst'
+import YAML from 'yaml'
+const parseCST = YAML.parseCST
```

This simplifies the public API of the library, but should not adversely affect users. Providing a separate endpoint has previously made it simpler for browser bundlers to treeshake the library for users that only use the CST parser, but the updated exports should now allow for that to happen even with a single export. Especially once #208 is resolved, as it should now be possible to provide ES-only exports for browsers.

This is also related to #203, which will change the output & API of `YAML.parseCST` rather thoroughly.


ssbarnea:
@eemeli While I see this as part of v2 release notes, there is nothing named `parseCST` in the current main branch and I see not PR referring this one. Any hints?

PR #227, opened 2021-02-05
Add option setOrigRanges for parsing docs

1nVitr0:
This pull request adds the following parameter to the document parser:

    setOrigRanges?: boolean

It uses the same name as the corresponding function `setOrigRanges`, which sets original ranges on cstNodes for `\r\n` style line endings. Default is `false` and it obviously only has an effect when `keepCstNodes` is set to `true`, although it does get executed either way, just that the CST nodes get discarded when `keepCstNodes` is set to `false`.

I realize adding more options should always be considered carefully, but being able to set the original ranges when parsing a document was pretty useful for me.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -118,6 +118,13 @@ export interface Options extends Schema.Options {
    */
   keepCstNodes?: boolean
   /**
+   * Set original ranges (support \r\n line endings) for included cstNodes
+   * Only has an effect if `keepCstNodes` is enabled
+   *
+   * Default: `false`
+   */
+  setOrigRanges?: boolean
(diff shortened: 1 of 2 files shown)


eemeli:
I'm going to hold off on really considering this until #203 gets merged, because that might remove the need for this: the new parser doesn't normalise the line endings. OTOH, atm it's only tracking the start indices of nodes, and I haven't figured out exactly yet if or how to provide end indices as well.

eemeli:
Closing, as CRLF newlines are no longer converted to LF at any parsing stage, and range positions are always valid for the original string.

PR #251, opened 2021-03-29
Ensure @node/types is available

fearthecowboy:
I cloned this and used PNPM instead of NPM and it doesn't find `@node/types` because it's not a devDependency.

(I assume it's installed as a dependency of something else, but PNPM isn't as lax as NPM when it installs things. )

- adds `@node/types`

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@rollup/plugin-replace": "^2.3.4",
     "@rollup/plugin-typescript": "^8.1.1",
     "@types/jest": "^26.0.20",
-    "@types/node": "14.14.22",
+    "@types/node": "*",
     "@typescript-eslint/eslint-plugin": "^4.15.2",
     "@typescript-eslint/parser": "^4.15.2",
     "babel-jest": "^26.6.3",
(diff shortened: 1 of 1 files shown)


eemeli:
That's rather surprising. I just did a fresh clone of the master branch, and at least for me `pnpm it` and `pnpm run test:types` succeed without any issue, and the transitive `@types/node` dependency does show up in `node_modules`.

Could you add instructions for reaching the error case that prompted this PR?

fearthecowboy:
Oh, my apologies. I assumed it was PNPM. 

My case had me adding your repo as a submodule to my monorepo project (managed with `@microsoft/rush` ) 
I guess then it's rush doing that.

(ah yes, I see that's what caused it. Hmm. I thought it was PNPM that was pedantic about that, turns out it's rush. )

Oh, and you're using `*` instead of a specific version.

eemeli:
Even though this smells like it might be a `@microsoft/rush` bug, there are a few places in the sources that do admittedly refer to Node.js types, so adding the dev dependency is valid. Done now, including an update to the lockfile.

PR #259, opened 2021-04-17
Refactor Node range as [start, value-end, node-end]

### LINKED ISSUE #258: In correct range
**Describe the bug**
The range for node corresponding to the end of a text is in correct and spans till the end of the text. 

**To Reproduce**
Using the text  `"definitions:\n  link:\n    type: string\ntype: object\nproperties:\n  uri:\n    $ref: '#/definitions/link'   \n"` I see that the range of single quote string of `#/definitions/link` extends to the whitespace while value is correct

**Expected behaviour**
The range of the string node not to go beyond

_next_ on vscode

eemeli:
Fixes #258 by adding a third value to the `range` tuple.

For block values, `value-end` and `node-end` are always the same as the previous end value. For flow collections and documents, `node-end` is the same as the previous `end` position, while `value-end` marks the end of the actual value.

@gorkem, would this work for your use case?

### DIFF
--- a/docs/04_documents.md
+++ b/docs/04_documents.md
@@ -14,33 +14,30 @@ doc.contents
 // YAMLMap {
 //   items:
 //    [ Pair {
-//        key: Scalar { value: 'YAML', range: [ 0, 4 ] },
+//        key: Scalar { value: 'YAML', range: [ 0, 4, 4 ] },
 //        value:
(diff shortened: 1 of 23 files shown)


eemeli:
Merging, as otherwise work on error ranges gets really clumsy.

PR #278, opened 2021-05-31
Fix empty lines & trailing comments

### LINKED ISSUE #244: Trailing comments are not well assigned
**Describe the bug**
Some trailing comments don't get assigned to the best possible node during parsing & composition.

**To Reproduce**
Un-`skip` [this section](https://github.com/eemeli/yaml/blob/59398d2d4e3d8c33a8131a04c34c17f518d0bd70/tests/doc/comments.js#L730) of the test suite, and observe the results, in particular for the `1` comments.

**Expected behaviour**
Either these comments should get assigned the same was as they did with the old parser, or we should have a really good story for why they ought to be handled differently.

- `yaml`: 2.0.0-4

**Additional context**
This is tough. The YAML spec says nothing about comments, so the only real yardstick is what "feels right". It's also tricky to handle at the end of a collection, when any number of surrounding collections might also be ending, and you can't tell until the next actual value which all of those ought to even be considered as the place to assign a trailing comment.

### LINKED ISSUE #277: Comment on the last line is moved on the line above
Considering this yml **input**
```
environment:
  ### SESSION START ###
  A: "true"
  B: "true"
  ### SESSION END ###
```

if the document is parsed and then printed out the result is not the same, but the comment goes on the same line as _'B'_.

```
const file = fs.readFileSync('./test/input/config.yml', 'utf8')
const doc = YAML.parseDocument(file)
doc.toString()
```

**output**
```
environment:
  ### SESSION START ###
  A: "true"
  B: "true" ### SESSION END ###
```

**Expected behaviour**
I would like to preserve the positioning of the comment. There is a way to achieve it?

**Versions:**

- Environment: 
- Node.js: v12.14.0
- yaml: 2.0.0-5

eemeli:
Fixes #244
Fixes #277 

I think I finally figured out a way to make comments and empty lines more usable. This hinges on refactoring the representation of completely empty lines vs. comment lines that terminate immediately after the `#`.

At the moment, empty lines in source are parsed into `spaceBefore` boolean values or completely dropped, and comments consisting of just a `#` are represented by an empty line in the `comment` or `commentBefore` value.

However, let's change that a bit. With this PR, an empty line in the source is represented by an empty line in the `comment` or `commentBefore` value, and a bare `#` is represented by a line consisting of a single space character. This relatively small change allows for any combination of comment and blank lines to be cleanly represented by a single string, with the cost of a comment consisting of a single space character being unrepresentable.

At least for now, an empty line before comments remains represented as the boolean `spaceBefore` property, but that property could be dropped entirely as it's equivalent to a `commentBefore` value of `'\n'`.

Simultaneously, the handling of trailing comments is simplified somewhat; they now need to be more indented than their parent collection to attach to the preceding rather than the following node.

### DIFF
--- a/docs/05_content_nodes.md
+++ b/docs/05_content_nodes.md
@@ -360,24 +360,25 @@ The return value of the visitor may be used to control the traversal:
 If `visitor` is a single function, it will be called with all values encountered in the tree, including e.g. `null` values.
 Alternatively, separate visitor functions may be defined for each `Map`, `Pair`, `Seq`, `Alias` and `Scalar` node.
 
-## Comments
+## Comments and Blank Lines
 
 ```js
 const doc = YAML.parseDocument(`
 # This is YAML.
 ---
 it has:
+
   - an array
+
   - of values
 `)
 
-doc.toJS()
-// { 'it has': [ 'an array', 'of values' ] }
+doc.toJS() // { 'it has': [ 'an array', 'of values' ] }
+doc.commentBefore // ' This is YAML.'
 
-doc.commentBefore
-// ' This is YAML.'
+const seq = doc.get('it has')
+seq.spaceBefore // true
 
-const seq = doc.contents.items[0].value
 seq.items[0].comment = ' item comment'
 seq.comment = ' collection end comment'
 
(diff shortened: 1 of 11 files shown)


gixlg:
Oh wow thanks for the awesome work!
This fix will be included in the next release (2.0.0-6)?

Issue #211, opened 2020-12-03
disable anchors in stringify


dudif91:
hi, is there a way to disable anchors such as &a1 when using stringify.

I'm using YAML version 1.10.0

tnx

eemeli:
I'm pretty sure that using just `YAML.stringify()` that isn't possible, but this will probably do what you want:

```js
const doc = new YAML.Document()
doc.setSchema()
doc.contents = doc.schema.createNode(yourValueHere)
String(doc)
```

The reason why that works is that in v1 the automatic reference detection was a part of the `stringify` method in particular, while `createNode` didn't include it. This changes (along with some of the clumsiness of the above API) in v2, but that's an entirely different issue.

dudif91:
tnx that was very helpful

ushuz:
> I'm pretty sure that using just `YAML.stringify()` that isn't possible, but this will probably do what you want:

Can this be made into an option of `stringify()`?

PR #299, opened 2021-08-12
Add a new createNode option aliasDuplicateObjects

### LINKED ISSUE #296: Add option to disable anchors in `stringify`
I'm using the stringify function to generate GitHub Actions workflows. The stringify function is creating anchors, which don't work with GitHub Actions.
> Anchors are not currently supported. Remove the anchor 'a1'

```js
import { stringify } from 'yaml'

stringify(someValue, { anchors: false })
```

- Deep cloning the js object so no anchors are created
- Following steps in #211 to use the solution.

**Additional context**
I'm creating this because I agree with @ushuz
> Can this be made into an option of stringify()?

https://github.com/eemeli/yaml/issues/211#issuecomment-812826265

eemeli:
Fixes #296, originally reported in #211

This adds an option `aliasDuplicateObjects` (default `true`) that does this:

```js
const foo = { foo: 'FOO' }

const doc1 = new YAML.Document([foo, foo])
const doc2 = new YAML.Document([foo, foo], { aliasDuplicateObjects: false })

`${doc1}
---
${doc2}`
```

```yaml
- &a1
  foo: FOO
- *a1

---
- foo: FOO
- foo: FOO
```

Ping @ChocolateLoverRaj, @ushuz, @dudif91. This would presumably work for you?

### DIFF
--- a/docs/03_options.md
+++ b/docs/03_options.md
@@ -99,12 +99,13 @@ Multiple merge keys may be used on the same map, with earlier values taking prec
 
 Used by: `stringify()`, `new Document()`, `doc.createNode()`, and `doc.createPair()`
 
-| Name          | Type      | Default | Description                                                                                                                                      |
-| ------------- | --------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
-| anchorPrefix  | `string`  | `'a'`   | Default prefix for anchors, resulting in anchors `a1`, `a2`, ... by default.                                                                     |
-| flow          | `boolean` | `false` | Force the top-level collection node to use flow style.                                                                                           |
-| keepUndefined | `boolean` | `false` | Keep `undefined` object values when creating mappings and return a Scalar node when stringifying `undefined`.                                    |
-| tag           | `string`  |         | Specify the top-level collection type, e.g. `"!!omap"`. Note that this requires the corresponding tag to be available in this document's schema. |
+| Name                  | Type      | Default | Description                                                                                                                                      |
+| --------------------- | --------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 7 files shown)


docs/03_options.md
```diff
@@ -99,12 +99,13 @@ Multiple merge keys may be used on the same map, with earlier values taking prec
...
-| flow          | `boolean` | `false` | Force the top-level collection node to use flow style.                                                                                           |
-| keepUndefined | `boolean` | `false` | Keep `undefined` object values when creating mappings and return a Scalar node when stringifying `undefined`.                                    |
-| tag           | `string`  |         | Specify the top-level collection type, e.g. `"!!omap"`. Note that this requires the corresponding tag to be available in this document's schema. |
+| Name                  | Type      | Default | Description                                                                                                                                      |
+| --------------------- | --------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| aliasDuplicateObjects | `boolean` | `true`  | During node construction, use anchors and aliases to keep strictly equal non-null objects as equivalent in YAML.                                 |
```
ushuz, line 104: `aliasDupplicateObjects` feels too implementation-focus. I would prefer something like `anchors` proposed by #211, which will sit well with `anchorPrefix` below and is easier for users to understand IMO.

ushuz (commented):
Thank you for addressing #211

ChocolateLoverRaj:
@eemeli whatever the option is called, it works for me.

derolf:
When will it be released?

eemeli:
@derolf https://github.com/eemeli/yaml/releases/tag/v2.0.0-8

GregKruz (approved):

eemeli:
> @eemeli Am I right in thinking this only works with duplicate objects within `Seq`uences, and not `Map`pings?

No, this works with all duplicate objects.

PR #309, opened 2021-09-11
Add keepSoureToken parse option, adding srcToken values to Nodes

### LINKED ISSUE #308: keepCstNodes was useful; consider restoring it

In 2.0.0-3, I could use the high-level `Node` API with `keepCstNodes` to easily inspect the document and select part of it to edit, then use the low-level `CST` API to edit that part and re-serialize the document without making unnecessary formatting changes to the other parts. However, in 2.0.0-4, the `keepCstNodes` option was removed, so this is no longer possible.

```js
import { parseDocument } from "yaml";

const yaml = `\
a: b
c:
    d: [e, f, g]
    h:
        {
            i: j,
            k: l,
        }
    m: n
o: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
`;

const doc = parseDocument(yaml, { keepCstNodes: true });
doc.contents.get("c").get("m", true).cstNode.value = "edited";
console.log(doc.cstNode.toString());

/*
a: b
c:
    d: [e, f, g]
    h:
        {
            i: j,
            k: l,
        }
    m: edited
o: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
*/
```

Restore the `keepCstNodes` option.

* Both `doc.toString()` and `stringify(doc)` make formatting changes to unrelated parts of the document.

  ```js
  doc.contents.get("c").set("m", "edited");
  console.log(doc.toString());

  /*
  a: b
  c:
    d: [ e, f, g ]
    h: { i: j, k: l }
    m: edited
  o: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
    tempor incididunt ut labore et dolore magna aliqua.
  */
  ```

* I could use `Parser.parse` and do the entire process using `CST` from start to finish, but that would require much more code for locating the part of the document I want to edit.

* I could use both `parseDocument` and `Parser.parse` and use techniques like the [`findScalarAtOffset` example](https://eemeli.org/yaml/#cst-stringify-cst-token-collectionitem-string) in the documentation to correspond AST nodes to CST nodes, but that would also require much more code.

eemeli:
Fixes #308 

This allows for usage like this:

```js
import { Composer, CST, Parser } from 'yaml'

const src = 'foo:\n  [ 42 ]'
const tokens = Array.from(new Parser().parse(src))

const docs = new Composer({ keepSourceTokens: true }).compose(tokens)
for (const doc of docs) {
  const node = doc.get('foo', true)
  CST.setScalarValue(node.srcToken, 'eek')
}

const res = tokens.map(CST.stringify).join('')
assert.equal(res, 'foo:\n  eek')
```

The `srcToken` value isn't available on nodes that are created from empty contents, as those do not have a matching CST token. This means that for example the above would fail if the `src` was `'foo:\n'`, as then `node.srcToken` is undefined.

To handle that case, you'd probably need to `CST.visit` the parent collection, identify the right CollectionItem, and then use `CST.createScalarToken()` when setting its `value`. At which point it might be more straightforward to keep all of the operations in the CST, and to never even compose it into documents.

### DIFF
--- a/docs/03_options.md
+++ b/docs/03_options.md
@@ -22,13 +22,14 @@ Parse options affect the parsing and composition of a YAML Document from it sour
 
 Used by: `parse()`, `parseDocument()`, `parseAllDocuments()`, `new Composer()`, and `new Document()`
 
-| Name         | Type                          | Default | Description                                                                                                                                                                |
-| ------------ | ----------------------------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| intAsBigInt  | `boolean`                     | `false` | Whether integers should be parsed into [BigInt] rather than `number` values.                                                                                               |
-| lineCounter  | `LineCounter`                 |         | If set, newlines will be tracked, to allow for `lineCounter.linePos(offset)` to provide the `{ line, col }` positions within the input.                                    |
-| prettyErrors | `boolean`                     | `true`  | Include line/col position in errors, along with an extract of the source string.                                                                                           |
-| strict       | `boolean`                     | `true`  | When parsing, do not ignore errors required by the YAML 1.2 spec, but caused by unambiguous content.                                                                       |
-| uniqueKeys   | `boolean ⎮ (a, b) => boolean` | `true`  | Whether key uniqueness is checked, or customised. If set to be a function, it will be passed two parsed nodes and should return a boolean value indicating their equality. |
+| Name             | Type                          | Default | Description                                                                                                                                                                |
+| ---------------- | ----------------------------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| intAsBigInt      | `boolean`                     | `false` | Whether integers should be parsed into [BigInt] rather than `number` values.                                                                                               |
+| keepSourceTokens | `boolean`                     | `false` | Include a `srcToken` value on each parsed `Node`, containing the CST token that was composed into this node.                                                               |
+| lineCounter      | `LineCounter`                 |         | If set, newlines will be tracked, to allow for `lineCounter.linePos(offset)` to provide the `{ line, col }` positions within the input.                                    |
+| prettyErrors     | `boolean`                     | `true`  | Include line/col position in errors, along with an extract of the source string.                                                                                           |
+| strict           | `boolean`                     | `true`  | When parsing, do not ignore errors required by the YAML 1.2 spec, but caused by unambiguous content.                                                                       |
+| uniqueKeys       | `boolean ⎮ (a, b) => boolean` | `true`  | Whether key uniqueness is checked, or customised. If set to be a function, it will be passed two parsed nodes and should return a boolean value indicating their equality. |
 
 [bigint]: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/BigInt
 
--- a/src/compose/compose-node.ts
+++ b/src/compose/compose-node.ts
@@ -68,6 +68,7 @@ export function composeNode(
     if (token.type === 'scalar' && token.source === '') node.comment = comment
     else node.commentBefore = comment
   }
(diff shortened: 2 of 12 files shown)


andersk:
> This means that for example the above would fail if the `src` was `'foo:\n'`, as then `node.srcToken` is undefined.

Might it make sense to use an empty pseudotoken for that case instead of `undefined`?

Or provide an API like

```ts
setKeyScalarValue(item: CollectionItem, value: string, context?);
setValueScalarValue(item: CollectionItem, value: string, context?);
```

for setting `item.key` and `item.value`, which would work even if they were initially undefined?

eemeli:
> > This means that for example the above would fail if the `src` was `'foo:\n'`, as then `node.srcToken` is undefined.
> 
> Might it make sense to use an empty pseudotoken for that case instead of `undefined`?

The problem with that approach is that in this case the source of the Node is not just `undefined`, it is literally undefined, as in there's no sequence of characters in the source for the Node. It is instead constructed from that very lack of characters.

> Or provide an API like
> 
> ```ts
> setKeyScalarValue(item: CollectionItem, value: string, context?): void;
> setValueScalarValue(item: CollectionItem, value: string, context?): void;
> ```
> 
> for setting `item.key` and `item.value`, which would work even if they were initially undefined?
> 
> Or an API that returns the a new `Token` rather than mutating the existing one, making it the user’s responsibility to insert it into the document?
> 
> ```ts
> makeScalarValue(token: Token | undefined, value: string, context?): Token;
> ```

I don't think these are necessary, as they'd end up as rather minimal wrappers around `CST.createScalarToken` and `CST.setScalarValue`.

> For the last two ideas, the user needs access to the `CollectionItem`, so a `YAMLMap.getPair(key: K): Pair | undefined` and a `Pair.srcToken?: CollectionItem` would be helpful.

The accessor you're looking for is `findPair(pairs, key)`, which is available from `'yaml/util'`. It should probably be better documented or removed. You can also iterate the pairs with `visit()`, which would give you similar results.

Adding `srcToken` also to Pair makes sense, and would provide a decent workaround for the `undefined` issues.

andersk:
> The problem with that approach is that in this case the source of the Node is not just `undefined`, it is literally undefined, as in there's no sequence of characters in the source for the Node. It is instead constructed from that very lack of characters.

An empty sequence is still a sequence. The suggestion was that you could invent a “token” corresponding to an empty string of source characters: `{ type: "null", offset: (next to the colon), source: "" }`.

> I don't think these are necessary, as they'd end up as rather minimal wrappers around `CST.createScalarToken` and `CST.setScalarValue`.

My goal was to find an API that would automatically work for the `undefined` case without requiring the user to explicitly branch for it as currently required:

```js
if (item.value === undefined)
    item.value = CST.createScalarToken(value, context);
else
    CST.setScalarValue(item.value, value, context);
```

But if you’re content with the existing API, don’t let me stop you from merging this.

eemeli:
If you can find a clean way of adding appropriate empty value tokens into the CST during the parse, then I might be interested in considering a PR for it. However, even that won't solve all cases, as explicit `? keys` don't need a matching `: value`.

Assigning a `srcToken` value that's unconnected to the actual source CST tree doesn't really make sense, as then modifying that would have no effect on the source CST. Or did I misunderstand your idea here?

Similarly, I'd be happy to consider a PR for `CST.setScalarValue` to be able to handle its first argument being empty, which would probably achieve what you're looking for

andersk:
Right, the first idea was that the parser would insert these empty tokens into the CST tree.

Without an object to represent the empty token, it’s not possible for `CST.setScalarValue` to handle its first argument being empty, for the same reason: modifying `undefined` would have no effect on the source CST. Hence the other two ideas—the `CollectionItem` is an object that can be passed to a function and mutated even when its `key` or `value` cannot.

GregKruz (approved):

PR #317, opened 2021-09-29
add `prepare` script to package.json so that `npm install`ing from a git commit works

cemerick:


cemerick:
According to [this](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts), dev dependencies _are_ installed anytime a `prepare` script is defined, so I don't think it would fail for lack of those. However, `src` is not included in the packages you are publishing, so yeah, things wouldn't get very far.

After poking around a bit (typescript & npm is definitely not my native territory), it seems this scenario — publishing a selective/derivative package to npm after a build process and ensuring that people can `npm install` from git sources, all while using the same `package.json` — is sort of an open problem (see SO discourse on the subject e.g. [here](https://stackoverflow.com/questions/51078974/how-to-have-npm-install-a-typescript-dependency-from-a-github-url) and [here](https://stackoverflow.com/questions/48287776/automatically-build-npm-module-on-install-from-github)).

There are some reasonable approaches, but they'd all require changes to your existing workflow or nontrivial rearrangement of the packaging approach. I'll just keep my own fork with the prepare in place and refer to that until the next release. :smile:

PR #344, opened 2021-12-27
Allow for Schema instance as schema option

eemeli:
Previously, `doc.setSchema` always created a new Schema instance based on the given schema identifier. This change allows for an externally defined schema to be used instead, which makes using a different predefined set of default option values easier.

### DIFF
--- a/src/doc/Document.ts
+++ b/src/doc/Document.ts
@@ -347,7 +347,7 @@ export class Document<T = unknown> {
 
   /**
    * Change the YAML version and schema used by the document.
-   * A `null` version disables support for directives and explicit tags.
+   * A `null` version disables support for directives, explicit tags, anchors, and aliases.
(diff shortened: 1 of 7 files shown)


eemeli:
Modified the PR a bit, such that the API now becomes:

```ts
type SchemaOptions = { ...
  /** The base schema to use. */
  schema?: string | Schema
}

class Document { ...
  directives?: Directives // previously required

  /**
   * Change the YAML version and schema used by the document.
   * A `null` version disables support for directives, explicit tags, anchors, and aliases.
   * It also requires the `schema` option to be given as a `Schema` instance value.
   *
   * Overrides all previously set schema options.
   */
  setSchema(version: '1.1' | '1.2' | null, options: SchemaOptions = {}): void
}
```

GregKruz (approved):

PR #371, opened 2022-03-15
Add directives.docEnd, for ... marker

### LINKED ISSUE #368: The ToString function removes the YAML document end marker if present in the document being parsed
**Describe the bug**

The `ToString` function removes the [YAML document end marker (`...`)](https://yaml.org/spec/1.2.2/#22-structures) if it's present in the document for which the `ToString` function is called.

I see that there's an option called `directivesEndMarker` that does the following, according to [docs](https://eemeli.org/yaml/v1/#creating-documents)):

> Whether the document should always include a directives-end marker `---` at its start, even if it includes no directives.

But that doesn't prevent the ToString function to strip the YAML document end marker. It also has a surprising (at least to me) name, because the `---` is the YAML document start, not the end.

I encountered this issue while investigating https://github.com/microsoft/compose-language-service/issues/93.

**To Reproduce**

Create the following YAML file:

```
---
services:
  test-service-1:
...
```

Load it, and call the ToString function on the resulting object, as done [here](https://github.com/microsoft/compose-language-service/blob/71abdc45d06f6f19af47e06df59168794675c345/src/service/providers/DocumentFormattingProvider.ts#L18):

```
        const options: ToStringOptions = {
            indent: params.options.tabSize,
            indentSeq: true,
            simpleKeys: true, // todo?
            nullStr: '',
            lineWidth: 0,
        };

        const range = Range.create(
            params.document.textDocument.positionAt(0),
            params.document.textDocument.positionAt(params.document.textDocument.getText().length)
        );

        const formatted = params.document.yamlDocument.value.toString(options);
```

This results in:

```
---
services:
  test-service-1:
```

**Expected behaviour**

The `ToString` shouldn't strip the YAML document end marker if present in the document being formatted or, at least, make this behavior configurable with an option.

- Environment: https://github.com/microsoft/compose-language-service
- `yaml`: "^2.0.0-10"

**Additional context**

/cc @bwateratmsft

eemeli:
Fixes #368, CC @ferrarimarco

This PR contains a **BREAKING CHANGE**, as it renames the `directives.marker` to `directives.docStart`. The behaviour of the property stays the same, though.

Next to that, we add `directives.docEnd` to mark whether the document contains a `...` doc-end marker.

### DIFF
--- a/docs/04_documents.md
+++ b/docs/04_documents.md
@@ -159,7 +159,7 @@ See [Options](#options) for more information on the optional parameter.
 const doc = new Document()
 doc.directives
 > {
-    marker: null, // set true to force the doc-start marker
+    docStart: null, // set true to force the doc-start marker
     tags: { '!!': 'tag:yaml.org,2002:' }, // Record<handle, prefix>
     yaml: { explicit: false, version: '1.2' }
   }
--- a/docs/04_documents.md
+++ b/docs/04_documents.md
@@ -160,6 +160,7 @@ const doc = new Document()
 doc.directives
 > {
     docStart: null, // set true to force the doc-start marker
+    docEnd: false, // set true to force the doc-end marker
     tags: { '!!': 'tag:yaml.org,2002:' }, // Record<handle, prefix>
     yaml: { explicit: false, version: '1.2' }
   }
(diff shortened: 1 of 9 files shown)


ferrarimarco:
Thanks @eemeli !

/cc @bwateratmsft

GregKruz (approved):

PR #380, opened 2022-04-18
Allow for pathlike aliases

eemeli:
This enables the resolution of YAML looking like this:

```yaml
- &foo 
  bar:
    - 1
    - 2
    - 42
- *foo/bar/2
```
to resolve as:
```js
[
  { bar: [1, 2, 42] },
  42
]
```

In other words, if an alias contains a `/` character, it is resolved as follows:
1. If a preceding anchor matches the alias exactly, use that and resolve the value as previously.
2. Else, find the last longest preceding anchor which matches the start of the alias if it were followed by a `/` character.
3. Considering the node of that anchor as a root node, use the rest of the alias as a `/` separated path to the target node.
4. Provided that the target node is a scalar or a collection, resolve it as the alias's value.

Put together, this means that all existing valid YAML continues to have the exact same meaning as before, but it's possible for an alias to point at nodes within a structure with an anchor. As with anchors, JS equality is maintained across aliased values, so e.g. this works:

```js
import { parse } from 'yaml'

const src = `
  - &foo { bar: [1, 2, 42] }
  - { alias: *foo/bar }
`
const res = parse(src, { version: 'next' })

Array.isArray(res[0].bar) // true
res[0].bar === res[1].alias // true
```

In the implementation, this currently requires tracking some additional nodes during the native value composition, but a later optimisation should be able to reduce the impact of that.

This feature is not yet in the YAML spec, and so requires using `version: 'next'` as an option. Its implementation is also not guaranteed to follow semver.

### DIFF
--- a/src/doc/Document.ts
+++ b/src/doc/Document.ts
@@ -423,11 +423,14 @@ export class Document<T = unknown> {
       mapAsMap: mapAsMap === true,
       mapKeyWarned: false,
       maxAliasCount: typeof maxAliasCount === 'number' ? maxAliasCount : 100,
+      resolved: new WeakMap(),
       stringify
     }
     const res = toJS(this.contents, jsonArg ?? '', ctx)
-    if (typeof onAnchor === 'function')
-      for (const { count, res } of ctx.anchors.values()) onAnchor(res, count)
+    if (typeof onAnchor === 'function') {
+      for (const [node, { count }] of ctx.anchors)
+        onAnchor(ctx.resolved.get(node), count)
+    }
     return typeof reviver === 'function'
       ? applyReviver(reviver, { '': res }, '', res)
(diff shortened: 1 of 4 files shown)


ioggstream:
Even more difficult: `2` might indicate either an integer key or the third array item

```yaml
- &baz
  baz:
    2: an-integer
    "2": a-string
   ? [1, 2]
   : a-list
   ? [1, "2"]
   : another-list
- *baz/2     # an-integer
- *baz/"2"   # a-string
- *baz/[1,2]  # a-list
- *baz/[1,"2"] # another-list
```

iiuc libfyaml uses `"` do discriminate between integers and strings.

Maybe there are other corner cases ..

GregKruz (approved):

Issue #382, opened 2022-04-20
Generic argument support for node functions


golergka:
`isMap` and similar methods should be able to accept generic parameter to more correctly specify the target type.

### Discussed in https://github.com/eemeli/yaml/discussions/381

<div type='discussions-op-text'>

<sup>Originally posted by **golergka** April 20, 2022</sup>
I'm using version 2.0.1. I know that this document should be a map, and I want to access it's property named "key". I tried doing it two different ways: both using primitive values and creating ndoes myself, [as described in documentation](https://eemeli.org/yaml/#collections), but both result in errors:

```
declare const doc: yaml.Document.Parsed;

if (yaml.isMap(doc.contents)) {

  // Argument of type 'string' is not assignable to parameter of type 'ParsedNode'
  doc.contents.get("key")

  // Argument of type 'Node' is not assignable to parameter of type 'ParsedNode'.
  doc.contents.get(doc.createNode("key"))
}
```

How am I supposed to access this map?</div>

PR #383, opened 2022-04-20
Generic argument support for node methods

golergka:
Closes https://github.com/eemeli/yaml/issues/382

I'd like to add some tests that use types as well, but the relevant test `collection-access.js` is currently untyped, and changing it seemed a little to large a change for a first contribution.

### DIFF
--- a/src/doc/Document.ts
+++ b/src/doc/Document.ts
@@ -193,7 +193,7 @@ export class Document<T extends Node = Node> {
    * Convert any value into a `Node` using the current schema, recursively
    * turning objects into collections.
    */
-  createNode<T = unknown>(value: T, options?: CreateNodeOptions): NodeType<T>
+  createNode<TValue = unknown>(value: TValue, options?: CreateNodeOptions): NodeType<TValue>
   createNode<T = unknown>(
     value: T,
     replacer: Replacer | CreateNodeOptions | null,
--- a/src/doc/Document.ts
+++ b/src/doc/Document.ts
@@ -264,7 +264,7 @@ export class Document<T = unknown> {
    * Removes a value from the document.
    * @returns `true` if the item was found and removed.
    */
-  delete(key: any) {
+  delete(key: unknown): boolean {
     return assertCollection(this.contents) ? this.contents.delete(key) : false
   }
 
@@ -272,7 +272,7 @@ export class Document<T = unknown> {
    * Removes a value from the document.
    * @returns `true` if the item was found and removed.
    */
-  deleteIn(path: Iterable<unknown>) {
+  deleteIn(path: Iterable<unknown> | null): boolean {
     if (isEmptyPath(path)) {
       if (this.contents == null) return false
       this.contents = null
@@ -288,7 +288,7 @@ export class Document<T = unknown> {
    * scalar values from their surrounding node; to disable set `keepScalar` to
    * `true` (collections are always returned intact).
    */
-  get(key: unknown, keepScalar?: boolean) {
+  get(key: unknown, keepScalar?: boolean): unknown {
     return isCollection(this.contents)
       ? this.contents.get(key, keepScalar)
       : undefined
@@ -299,7 +299,7 @@ export class Document<T = unknown> {
    * scalar values from their surrounding node; to disable set `keepScalar` to
    * `true` (collections are always returned intact).
    */
-  getIn(path: Iterable<unknown>, keepScalar?: boolean) {
+  getIn(path: Iterable<unknown> | null, keepScalar?: boolean): unknown {
     if (isEmptyPath(path))
       return !keepScalar && isScalar(this.contents)
         ? this.contents.value
@@ -312,14 +312,14 @@ export class Document<T = unknown> {
   /**
    * Checks if the document includes a value with the key `key`.
    */
-  has(key: unknown) {
+  has(key: unknown): boolean {
     return isCollection(this.contents) ? this.contents.has(key) : false
   }
 
   /**
    * Checks if the document includes a value at `path`.
    */
-  hasIn(path: Iterable<unknown>) {
+  hasIn(path: Iterable<unknown> | null): boolean {
     if (isEmptyPath(path)) return this.contents !== undefined
     return isCollection(this.contents) ? this.contents.hasIn(path) : false
   }
@@ -328,7 +328,7 @@ export class Document<T = unknown> {
    * Sets a value in this document. For `!!set`, `value` needs to be a
    * boolean to add/remove the item from the set.
    */
-  set(key: any, value: unknown) {
+  set(key: any, value: unknown): void {
     if (this.contents == null) {
       this.contents = collectionFromPath(
         this.schema,
@@ -344,7 +344,7 @@ export class Document<T = unknown> {
    * Sets a value in this document. For `!!set`, `value` needs to be a
    * boolean to add/remove the item from the set.
    */
-  setIn(path: Iterable<unknown>, value: unknown) {
+  setIn(path: Iterable<unknown> | null, value: unknown): void {
     if (isEmptyPath(path)) this.contents = value as T
     else if (this.contents == null) {
       this.contents = collectionFromPath(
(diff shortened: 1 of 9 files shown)


src/nodes/Node.ts
```diff
@@ -7,7 +7,11 @@ import type { Scalar } from './Scalar.js'
 import type { YAMLMap } from './YAMLMap.js'
 import type { YAMLSeq } from './YAMLSeq.js'
 
-export type Node = Alias | Scalar | YAMLMap | YAMLSeq
+export type Node<T = unknown> =
```
golergka, line 10: May be it would make more sense to rename `K` to `Key` and both `T` and `V` to `Value` across the whole codebase? Because now the fact that `T` and `V` have the same meaning (type of the value) may not be immediately obvious for the reader.

eemeli: Yeah, probably. The generic type identifiers are a bit terse.

eemeli:
@golergka I'd like to expand this PR a bit to package in a few more TS changes, including a port of `collection-access.js` to TypeScript. Would you be okay with me adding commits to your branch? That should allow you to also add tests for the current changes.

golergka:
> @golergka I'd like to expand this PR a bit to package in a few more TS changes, including a port of `collection-access.js` to TypeScript. Would you be okay with me adding commits to your branch? That should allow you to also add tests for the current changes.

Certainly! It's your repo, after all 😁

src/index.ts
```diff
@@ -48,6 +48,8 @@ export {
 
 export type { TagId, Tags } from './schema/tags'
 export type { CollectionTag, ScalarTag } from './schema/types'
+export type { YAMLOMap } from './schema/yaml-1.1/omap'
+export type { YAMLSet } from './schema/yaml-1.1/set'
```
eemeli, line 52: As these were needed for the collection-access tests, they may be useful to others as well.

src/nodes/Node.ts
```diff
@@ -7,7 +7,20 @@ import type { Scalar } from './Scalar.js'
...
+  | YAMLSeq<T>
+
+/** Utility type mapper */
+export type NodeType<T> = T extends string | number | bigint | boolean | null
+  ? Scalar<T>
+  : T extends Array<any>
+  ? YAMLSeq<NodeType<T[number]>>
+  : T extends { [key: string | number]: any }
+  ? YAMLMap<NodeType<keyof T>, NodeType<T[keyof T]>>
+  : Node
```
eemeli, line 23: This isn't currently publicly exported, but I wonder if it'd be useful outside of `createNode()`? Also happy to bikeshed on the name.

src/doc/Document.ts
```diff
@@ -41,7 +42,7 @@ export declare namespace Document {
   }
 }
 
-export class Document<T = unknown> {
+export class Document<T extends Node = Node> {
```
eemeli, line 45: I would _really_ like for `new Document(foo)` to automatically type itself as `Document<NodeType<typeof foo>>`, but I can't figure out how to make that happen.

eemeli (changes requested):
@golergka Please re-request a review from me once you've added tests for your part and had a look at my additions? I've incorporated here changes that should resolve the issues from #381 as well.

src/nodes/YAMLMap.ts
```diff
@@ -76,24 +76,30 @@ export class YAMLMap<K = unknown, V = unknown> extends Collection {
...
-  delete(key: K) {
+  delete(key: unknown): boolean {
     const it = findPair(this.items, key)
     if (!it) return false
     const del = this.items.splice(this.items.indexOf(it), 1)
     return del.length > 0
   }
 
-  get(key: K, keepScalar?: boolean) {
+  get(key: unknown, keepScalar: true): Scalar<V> | undefined
```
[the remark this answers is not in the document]
eemeli: Nope, you're right, not making these generic is the right call. The `is*` function generics may be used to do that as a side effect, but that too makes sense.

src/doc/Document.ts
```diff
@@ -192,12 +193,12 @@ export class Document<T = unknown> {
    * Convert any value into a `Node` using the current schema, recursively
    * turning objects into collections.
    */
-  createNode(value: unknown, options?: CreateNodeOptions): Node
-  createNode(
-    value: unknown,
+  createNode<TValue = unknown>(value: TValue, options?: CreateNodeOptions): NodeType<TValue>
+  createNode<T = unknown>(
+    value: T,
```
eemeli, line 198: Would be good to keep the same generic type identifier across function signatures.

src/nodes/Node.ts
```diff
@@ -7,7 +7,20 @@ import type { Scalar } from './Scalar.js'
...
+  | YAMLSeq<T>
+
+/** Utility type mapper */
+export type NodeType<TValue> = TValue extends string | number | bigint | boolean | null
+  ? Scalar<TValue>
+  : TValue extends Array<any>
+  ? YAMLSeq<NodeType<TValue[number]>>
+  : TValue extends { [key: string | number]: any }
+  ? YAMLMap<NodeType<keyof TValue>, NodeType<TValue[keyof TValue]>>
+  : Node
```
eemeli, line 23: Any particular reason to prefer `TValue` rather than something possibly simpler like `Value`?

src/nodes/YAMLMap.ts
```diff
@@ -76,24 +76,30 @@ export class YAMLMap<K = unknown, V = unknown> extends Collection {
...
+  get(key: unknown, keepScalar?: boolean): V | undefined
```
eemeli, line 87:
```suggestion
  get(key: unknown, keepScalar?: false): V | undefined
```
The general `boolean` case is caught in the next overload signature.

src/nodes/YAMLSeq.ts
```diff
@@ -56,11 +56,16 @@ export class YAMLSeq<T = unknown> extends Collection {
    * `key` must contain a representation of an integer for this to succeed.
    * It may be wrapped in a `Scalar`.
    */
-  get(key: unknown, keepScalar?: boolean) {
+  get(key: unknown, keepScalar: true): Scalar<T> | undefined
+  get(key: unknown, keepScalar?: boolean): T | undefined
```
eemeli, line 60:
```suggestion
  get(key: unknown, keepScalar?: false): T | undefined
```

tests/collection-access.ts
```diff
@@ -38,14 +59,25 @@ describe('Map', () => {
...
     expect(map.get('a', true)).toMatchObject({ value: 1 })
-    expect(map.get('b').toJSON()).toMatchObject({ c: 3, d: 4 })
+    const subMap = map.get('b')
+    if (!isMap<string, number>(subMap)) {
+      throw new Error('Expected subMap to be a Map')
+    }
+    expect(subMap.toJSON()).toMatchObject({
+      c: 3,
+      d: 4
+    })
```
eemeli, line 69:
```suggestion
    if (isMap<string, number>(subMap)) {
      expect(subMap.toJSON()).toMatchObject({ c: 3, d: 4 })
    } else {
      throw new Error('Expected subMap to be a Map')
    }
```

I like this style of assertion + cast. For readability, I would prefer the throws to come from explicit rather than implicit branches, though. Same goes for later tests as well.

eemeli (commented):
I think this is getting good. A few nitpicky fixes, and then we can merge this.

src/doc/Document.ts
golergka: Sorry, it was a part of attempted generic parameter rename that I committed by mistake. Reversed.

src/nodes/Node.ts
golergka: `Value` is already used in other places of the codebase, so I thought it would make sense to distinguish the generic parameters — would make relevant find and rename operations easier.

eemeli (approved):

eemeli:
Thank you @golergka! Very happy to have this included. I've a couple of other minor things that I want to get in, but then should be able to get a new release out within a week or so.

PR #396, opened 2022-06-03
Generate a package.json file in browser/dist

EricWittmann:
When using this module as a shared dependency in a webpack federated module project, a package.json file seems to be needed in browser **and** browser/dist **and** they both must have a `version` property.  Without this, the following error results:

```
Compiled with problems:

WARNING in shared module yaml -> C:\Users\Eric\git\api-designer\ads-ui\node_modules\yaml\browser\index.js

No version specified and unable to automatically determine one. No version in description file (usually package.json). Add version to description file, or manually specify version in shared config.
```

I've updated the build to generate these two `package.json` files.


eemeli:
This is odd, and sounds more like a bug in the tool you're using in your build. Would you be able to provide a reproduction of the original issue so this can be diagnosed further?

EricWittmann:
Sure.  I'll see if I can create as simple a reproducer as possible.

EricWittmann:
Reproducer here:  https://github.com/EricWittmann/module-federation-with-yaml-issue

The issue appears as only a warning during the build, but is more of a blocker when in dev mode.

eemeli:
This looks like an instance of webpack/webpack#13457 or something very close to it.

The warning is coming from [here](https://github.com/webpack/webpack/blob/b8748cfe149d21a71dbde0c4fa34d0646fe68cf8/lib/sharing/ProvideSharedPlugin.js#L131-L132). This is a Webpack bug as that plugin only looks at the nearest package.json file, rather than the nearest one which does define a `"name"` and a `"version"`, as it should.

I would recommend that you file a bug (and possibly a PR for a fix?) in Webpack for this.

EricWittmann:
OK I can try that.  I've seen other projects hit this issue and fix it in their own projects. But if it's a webpack bug it certainly makes more sense to attack it at the root.  I'll see what I can do.

EricWittmann:
This comment in webpack suggests a workaround (and that they may not be open to a resolution fix):

https://github.com/webpack/webpack/issues/13457#issuecomment-1017970025

I think I'll add a comment there with a reference to this PR (and reproducer) for context.  We'll also try the workaround when we have a chance...

vankop:
`name` and `version` are required fields in `package.json`.

miteokaku (approved):

miteokaku (approved):

eemeli:
Closing, as this needs to be fixed in Webpack: webpack/webpack#13457.

PR #407, opened 2022-08-25
Fix broken browserstack src

Quezler:
Came across https://deno.land/x/yaml@v2.1.1 when deciding on a config file format to use, noticed the issue & fixed it ^-^

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,5 +139,5 @@ YAML.stringify({ number: 3, plain: 'string', block: 'two\nlines\n' })
 Browser testing provided by:
 
 <a href="https://www.browserstack.com/open-source">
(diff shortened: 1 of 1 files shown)


eemeli:
Thank you for noticing this!

It looks like the docs-slate config that was supposed to disable asset hashing got overwritten at some point when merging updates from upstream.

So the right fix here was not to accept the hash, but to fix the config and re-publish the docs, as that got this fixed in previously deployed locations as well, such as [deno.land](https://deno.land/x/yaml@v2.1.1) and [npmjs.com](https://www.npmjs.com/package/yaml).

See the `yaml-docs` commit linked above for the change.

PR #418, opened 2022-10-25
[toJSON] empty string to non-null empty object.

rubenhak:
Added a flag which allows conversion of empty strings to an empty object, instead of returning null.

### DIFF
--- a/src/doc/Document.ts
+++ b/src/doc/Document.ts
@@ -414,6 +414,7 @@ export class Document<T extends Node = Node> {
     jsonArg,
     mapAsMap,
     maxAliasCount,
+    emptySourceAsObject,
     onAnchor,
     reviver
   }: ToJSOptions & { json?: boolean; jsonArg?: string | null } = {}): any {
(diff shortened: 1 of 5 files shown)


eemeli:
This would potentially be a bit problematic, as it's directly against the current YAML spec:

> [**7.2. Empty Nodes**](https://yaml.org/spec/1.2.2/#72-empty-nodes)
> YAML allows the node content to be omitted in many cases. Nodes with empty content are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a “`null`” value.

@rubenhak Could you describe what problem is being solved by this proposed change?

eemeli:
Closing due to lack of explanation.

rubenhak:
@eemeli, somehow missed your prior response. This is related to a few frameworks that generate Kubernetes Yaml manifests. They set some of the values 'null', expecting the backend to auto-populate the values. I understand it is against the YAML spec. Ended up implementing this hack in the app layer.

eemeli:
@rubenhak Glad that you got this sorted. Solving this at the application level sounds like it was the right fix.

PR #458, opened 2023-04-04
Update addPairToJSMap.ts, missing ? operator

MikeRalphson:
Found while testing alias preservation, using the `visit` method:

```js
function parseWithAliases(str) {
  const aliases = new Map();
  const ast = yaml.parseDocument(str);
  yaml.visit(ast,function(key,node,path) {
    if (yaml.isAlias(node)) {
      aliases.set(node.source,node.resolve(ast).toJS());
    }
  });
  return { data: ast.toJS(), aliases };
}
```

### DIFF
--- a/src/nodes/addPairToJSMap.ts
+++ b/src/nodes/addPairToJSMap.ts
@@ -13,7 +13,7 @@ export function addPairToJSMap(
   map: MapLike,
   { key, value }: Pair
 ) {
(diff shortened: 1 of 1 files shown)


eemeli:
Would you have some example that fails without this? At least internally `addPairToJSMap()` should never get called with a broken `ctx` value.

eemeli:
Closing with the fix committed in aa0a1d4.

PR #463, opened 2023-04-12
[package.json] add types in exports field

thibthib:
Starting from TS 4.7 ([see release note](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing)), we need to add a types field in the exports object when then path to the .d.ts file doesn’t match the file exposed.

If it isn’t set, we have the following error:

```
There are types at '[...]/node_modules/yaml/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The 'yaml' library may need to update its package.json or typings.
```

This should fix it by adding the same types as top level, but within the exports object

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,11 +25,13 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "node": "./dist/index.js",
       "default": "./browser/index.js"
     },
     "./package.json": "./package.json",
     "./util": {
+      "types": "./dist/util.d.ts",
       "node": "./dist/util.js",
       "default": "./browser/dist/util.js"
     }


tal-rofe:
@thibthib The issue occurs to me also. Is there WA in the mean time?

thibthib:
@tal-rofe The only workaround is to patch the library locally, and it will widely depend on the package manager you are using.

@eemeli Thanks for the review ! Do I need to do anything else to see this change in a new version of the library ?

Ayc0:
> patch the library locally, and it will widely depend on the package manager you are using.

With yarn 2+ you can use the `patch:` protocol ([link](https://yarnpkg.com/features/protocols/#patch)), with pnpm you can use `patchedDependencies` ([link](https://pnpm.io/package_json#pnpmpatcheddependencies)), and with npm you can use the package [patch-package](https://www.npmjs.com/package/patch-package)

eemeli:
> @eemeli Thanks for the review ! Do I need to do anything else to see this change in a new version of the library ?

No need for further action from you. I'll release a new prerelease version shortly.

eemeli:
There was an incompatibility with the `typesVersions` as well, which had been an attempt at fixing the types for the `'yaml/util'` endpoint. The types should now work as intended with `yaml@2.3.0-3`.

thibthib:
Thanks a lot @eemeli !

PR #494, opened 2023-08-26
Do not require quotes for implicit keys with flow indicators

### LINKED ISSUE #493: Option to plain-format some keys like /models/{model}
I'm trying to minimize diffs when using `yaml-diff-patch` on OpenAPI specs like
```yaml
openapi: 3.0.0
paths:
  /models/{model}:
    get:
      # etc
```

The `/models/{model}` key gets quoted after a roundtrip.

Question: is it actually a valid key?  It seems to parse okay and linting tools seem okay with it.  But, I found in your code that you intentionally quote implicit keys containing `{` or `}`, and the [YAML spec seems pretty clear on this](https://yaml.org/spec/1.2.2/#733-plain-style):

> In addition, inside [flow collections](https://yaml.org/spec/1.2.2/#flow-collection-styles), or when used as [implicit keys](https://yaml.org/spec/1.2.2/#example-single-pair-explicit-entry), plain scalars must not contain the “[”, “]”, “{”, “}” and “,” characters. These characters would cause ambiguity with [flow collection](https://yaml.org/spec/1.2.2/#flow-collection-styles) structures.

If an implicit key like `/models/{model}` is actually unambiguous, can we have `defaultKeyType: 'PLAIN'` not quote it?

If using an unquoted key in this form is bad and not technically valid, I'm wondering if there's a good example why it shouldn't be done I can take back to folks who are outputting YAML in this format?

**Additional context**

eemeli:
Fixes #493

The oddity here turned out to be in part caused by a mismatch in the YAML spec text & rules; I've filed yaml/yaml-spec#313 to add an errata for that. See there for more technical details, but the fix required here is removing one part of the check in the stringifier. I'd followed the YAML rules for the parser, but the YAML text for the stringifier, and the latter was wrong.

### DIFF
--- a/src/stringify/stringifyString.ts
+++ b/src/stringify/stringifyString.ts
@@ -280,7 +280,7 @@ function plainString(
   const { type, value } = item
   const { actualString, implicitKey, indent, indentStep, inFlow } = ctx
   if (
-    (implicitKey && /[\n[\]{},]/.test(value)) ||
+    (implicitKey && value.includes('\n')) ||
     (inFlow && /[[\]{},]/.test(value))
   ) {
     return quotedString(value, ctx)
--- a/tests/doc/stringify.ts
+++ b/tests/doc/stringify.ts
@@ -336,6 +336,11 @@ z:
     expect(String(doc)).toBe('a: 1\n? b: 2\n  c: 3\n')
   })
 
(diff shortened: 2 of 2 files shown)


rattrayalex:
Thank you!

PR #498, opened 2023-10-13
Fix malformed uri

manunio:
While [Fuzzing](https://en.wikipedia.org/wiki/Fuzzing) locally using [jazzer.js](https://github.com/CodeIntelligenceTesting/jazzer.js) `YAML.parse()` threw uncaught exception at
https://github.com/eemeli/yaml/blob/a4d8569c2139eee8a0324b6b861283f7b3b508f8/src/doc/directives.ts#L149 this pr attempts to fix that.
```
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at Directives.tagName (/home/maxx/dev/security/oss-fuzz-projects/eemeli-yaml/dist/doc/directives.js:126:29)
```

### DIFF
--- a/src/doc/directives.ts
+++ b/src/doc/directives.ts
@@ -149,8 +149,9 @@ export class Directives {
     if (prefix) {
       try {
         return prefix + decodeURIComponent(suffix)
-      } catch (_) {
-        onError('Failed to decode suffix')
+      } catch (error) {
+        onError(String(error))
+        return null
       }
     }
     if (handle === '!') return source // local tag
--- a/src/doc/directives.ts
+++ b/src/doc/directives.ts
@@ -149,7 +149,7 @@ export class Directives {
     if (prefix) {
       try {
         return prefix + decodeURIComponent(suffix)
-      } catch (error) {
+      } catch (_) {
         onError('Failed to decode suffix')
       }
     }
--- a/src/doc/directives.ts
+++ b/src/doc/directives.ts
@@ -146,7 +146,13 @@ export class Directives {
     const [, handle, suffix] = source.match(/^(.*!)([^!]*)$/) as string[]
     if (!suffix) onError(`The ${source} tag has no suffix`)
     const prefix = this.tags[handle]
-    if (prefix) return prefix + decodeURIComponent(suffix)
+    if (prefix) {
+      try {
+        return prefix + decodeURIComponent(suffix)
+      } catch (error) {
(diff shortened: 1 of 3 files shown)


src/doc/directives.ts
```diff
@@ -146,7 +146,13 @@ export class Directives {
     const [, handle, suffix] = source.match(/^(.*!)([^!]*)$/) as string[]
     if (!suffix) onError(`The ${source} tag has no suffix`)
     const prefix = this.tags[handle]
-    if (prefix) return prefix + decodeURIComponent(suffix)
+    if (prefix) {
+      try {
+        return prefix + decodeURIComponent(suffix)
+      } catch (_) {
+        onError('Failed to decode suffix')
```
eemeli, line 153: Is there a reason to not use the message of the error that's thrown here?

tests/doc/parse.ts
```diff
@@ -819,3 +819,10 @@ describe('CRLF line endings', () => {
     expect(res).toBe('foo bar')
   })
 })
+
+describe('Failed to decode suffix', () => {
+  test('rise in parse for malformed uri', () => {
+    const data = `[!!eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeJJJJJJJeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeebooleaeeeeeeeeeee-eeeeeeeee%eeeeeeeeeeeee[`
+    expect(() => YAML.parse(data)).toThrow(/Failed to decode suffix/)
```
eemeli, line 826: A few notes here:
- the test should call `YAML.parseDocument`, which should never throw.
- the check should be on the `doc.errors` contents; see for example the duplicate keys tests above.
- the source should not be triggering any other errors, so something like `!!%ee 0`.

eemeli (changes requested):
Thank you for reporting this! In general the fix looks right, but see line comments for details.

src/doc/directives.ts
eemeli: Could go with just `onError(String(error))`.

tests/doc/parse.ts
eemeli: Yeah, that works.

eemeli (approved):
Looks good! I realised that there's a better place for the test, so I moved it there.

Issue #503, opened 2023-11-10
Reported type in vscode is possibly wrong


wtfacoconut:
**Describe the bug**
I'm currently using this library in a new SvelteKit project with TypeScript.
The easiest way for me to explain the bug is with a code snippet of my code:
```
function printParsedYaml(){
  if(yamlSrcCode){
	  let tmpObj: Document = parseDocument(yamlSrcCode); 
	  let tmpItems: YAMLMap = tmpObj.contents; // vscode is wrong for some reason contents is of type YAMLmap, not Node. I checked the constructor name in chrome console, and it's YAMLMap
	  console.log(tmpObj.contents.items); // Because of the above issue, vscode things the items property doesn't exist
	  console.log(tmpItems.items);
  } else{
	  console.log("No src to parse.");
  }
}
```

**To Reproduce**
Please see the code snippet above

**Expected behaviour**
That vscode would report the correct return type

**Versions (please complete the following information):**
- Node.js: 18.17.1
- yaml: 2.3.3
- vscode: 1.84..0

eemeli:
How do you propose for the TS tooling to know that your YAML source represents a map?

wtfacoconut:
Hey @eemeli. I'm new to using this library (and new to JS/TS), so thanks for the speedy reply.
Based on you response, I'm guessing for any given node, the library does not determin the node's type (eg: YAMLMap, YAMLSeq, Scalar or null) until it is parsed at runtime. Is my understanding correct? 

I guess because I didn't notice anything with the feedback from intellisense, this went over my head. 
![image](https://github.com/eemeli/yaml/assets/5297545/561a1692-f1b2-4b61-865a-dbfd7b92925b)

Maybe the description for the contents property could be updated to say something like:
```
The document contents. At runtime, the document node type will be of type YAMLMap, YAMLSeq, Scalar or null.
```

The only section in the library's documentation that looks like it mentions this is at the bottom of the section for `creating documents` [https://eemeli.org/yaml/#creating-documents](link)
![image](https://github.com/eemeli/yaml/assets/5297545/60d09de1-2bb6-433e-9c0a-9d53316cf4e5)

Again, I'm new to using this library and I'm not knowledgable when it comes to things like parsers, so thanks for bearing with me. I highly suspect what I'm pointing out is a non-issue and is just something that people who typically use parsers in the JS/TS world understand as a convention (or something along those lines).

eemeli:
[...]

PR #504, opened 2023-11-11
Update Document.ts

wtfacoconut:
The update to the documentation is mainly for TS so that people are more aware that a node's type isn't determined until runtime.
This is related to issue #503 .


eemeli:
See https://github.com/eemeli/yaml/issues/503#issuecomment-1806585949

Issue #507, opened 2023-12-03
Flow collections always wrap at >80 characters, regardless of lineWidth option


TheBITLINK:
**Describe the bug**
When using the `flow` collection style, collections always seem to wrap at around 80 characters, regardless of the `lineWidth` option:

```yaml
blockCollection:
  aKey: hello
  anotherKey: world
  yetAnotherKey: "!!!"
listOfFlowCollection:
  - {
      aKey: hello,
      anotherKey: world,
      yetAnotherKey: "!!!",
      oneMoreKey: "!!!!!!!!"
    }
  - {
      aKey: hello,
      anotherKey: world,
      yetAnotherKey: "!!!",
      oneMoreKey: "!!!!!!!!"
    }
flowColTwo:
  - { aKey: hello, bKey: world, cKey: "!!!" }
  - { aKey: hello, bKey: world, cKey: "!!!" }
  - { aKey: hello, bKey: world, cKey: "!!!", dKey: "!" }
```

**To Reproduce**
```js
  function serialize() {
    const doc = new YAML.Document({
      blockCollection: {
        aKey: 'hello',
        anotherKey: 'world',
        yetAnotherKey: '!!!'
      },
      listOfFlowCollection: [
        { aKey: 'hello', anotherKey: 'world', yetAnotherKey: '!!!', oneMoreKey: '!!!!!!!!'  },
        { aKey: 'hello', anotherKey: 'world', yetAnotherKey: '!!!', oneMoreKey: '!!!!!!!!'  },
      ],
      flowColTwo: [
        { aKey: 'hello', bKey: 'world', cKey: '!!!'  },
        { aKey: 'hello', bKey: 'world', cKey: '!!!'  },
        { aKey: 'hello', bKey: 'world', cKey: '!!!',  dKey: '!' },
      ]
    })
    YAML.visit(doc, {
      Pair(_, pair: any) {
        if (pair.key?.value === 'listOfFlowCollection' || pair.key?.value === 'flowColTwo')  {
          for (const i of pair.value.items) {
            i.flow = true
          }
        }
      }
    })
    return doc.toString({ lineWidth: 120 })
  }
```

**Expected behaviour**
Flow collections respect the `lineWidth` option and wrap at 120 characters instead of 80 (or not wrap at all if `lineWidth` is 0)

```yaml
blockCollection:
  aKey: hello
  anotherKey: world
  yetAnotherKey: "!!!"
listOfFlowCollection:
[...]

Issue #288, opened 2021-06-17
Is there a way to change maxFlowStringSingleLineLength ?


eemeli:
### Discussed in https://github.com/eemeli/yaml/discussions/286

<div type='discussions-op-text'>

<sup>Originally posted by **codebot** June 17, 2021</sup>
Hello :wave:  Thank you for publishing this great library.

I have a small question:

I am trying to read-modify-write some existing YAML files without generating a giant diff, and would like to change `Collection.maxFlowStringSingleLineLength` from its hard-coded value of 60 to a larger number:

https://github.com/eemeli/yaml/blob/master/src/nodes/Collection.ts#L43

Is it possible to do this from user code?

Thank you.</div>

The intent was for this to be externally accessible, but it certainly doesn't look like it currently is.

eemeli:
I found a way, but it's not at all obvious:

```js
import { YAMLMap } from 'yaml'

Object.getPrototypeOf(YAMLMap).maxFlowStringSingleLineLength
```

Given that this is technically possible already now without API changes, I'm going to close this as `wontfix` for now, but I'm open to reconsidering that if this turns out to be of interest to other users.

codebot:
Thank you so much! This does exactly what I was hoping. Thank you for looking into this so quickly and for providing this suggestion. I would have never come up with that :rofl: Thanks again!

PR #522, opened 2024-02-12
fix: use the `lineWidth` option for line breaking in flow collections

no92:
This should address issues #288 and #507.

### DIFF
--- a/src/stringify/stringifyCollection.ts
+++ b/src/stringify/stringifyCollection.ts
@@ -148,7 +148,7 @@ function stringifyFlowCollection(
   } else {
     if (!reqNewline) {
       const len = lines.reduce((sum, line) => sum + line.length + 2, 2)
-      reqNewline = len > Collection.maxFlowStringSingleLineLength
+      reqNewline = ctx.options.lineWidth > 0 && len > ctx.options.lineWidth
     }
     if (reqNewline) {
       str = start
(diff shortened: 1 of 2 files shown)


src/stringify/stringifyCollection.ts
```diff
@@ -148,7 +148,7 @@ function stringifyFlowCollection(
   } else {
     if (!reqNewline) {
       const len = lines.reduce((sum, line) => sum + line.length + 2, 2)
-      reqNewline = len > Collection.maxFlowStringSingleLineLength
+      reqNewline = ctx.options.lineWidth !== 0 && len > ctx.options.lineWidth
```
eemeli, line 151: "Falsy" values for `lineWidth` include negative numbers. See these examples:

https://github.com/eemeli/yaml/blob/b7696fc001837a2e9d66ad78d7c04f47943daeca/src/stringify/foldFlowLines.ts#L53

https://github.com/eemeli/yaml/blob/b7696fc001837a2e9d66ad78d7c04f47943daeca/src/stringify/stringifyString.ts#L36

eemeli (changes requested):
Looks promising! Should include an additional test or three to exercise the option. Also, see inline comment.

src/stringify/stringifyCollection.ts
no92: Fixed by turning it into a `ctx.options.lineWidth > 0 && ...`

no92:
I also amended the commit to include two testcases, with one that wraps and another test that does not.

eemeli (approved):

PR #523, opened 2024-02-24
Add CLI tool

### LINKED ISSUE #479: Should a basic CLI be included?
I'm considering adding a command-line interface for `yaml`. Something pretty basic, and reliant on the Node.js [util.parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig) that was added in v16, and is no longer experimental in v20.

Is this something that you might find useful? If so, how minimal could it be in order to be useful? Reactions and comments here would be helpful.

eemeli:
Closes #479, CC @danielbayley & @antonmedv

Here's the `npx yaml -h` output, as currently implemented. ~I'm still contemplating a few additions like a node style normaliser, and the docs will need an update.~

```
yaml: A command-line YAML processor and inspector

Reads stdin and writes output to stdout and errors & warnings to stderr.

Usage:
  yaml          Process a YAML stream, outputting it as YAML
  yaml cst      Parse the CST of a YAML stream
  yaml lex      Parse the lexical tokens of a YAML stream
  yaml valid    Validate a YAML stream, returning 0 on success

Options:
  --help, -h    Show this message.
  --json, -j    Output JSON.

Additional options for bare "yaml" command:
  --doc, -d     Output pretty-printed JS Document objects.
  --single, -1  Require the input to consist of a single YAML document.
  --strict, -s  Stop on errors.
  --visit, -v   Apply a visitor to each document (requires a path to import)
  --yaml 1.1    Set the YAML version. (default: 1.2)
```


eemeli:
Added a `--visit` option. With that, you can define a [visitor](https://eemeli.org/yaml/#modifying-nodes) like:
```js
// strip-styles.mjs
export default {
  Collection(_, node) {
    delete node.flow
  },
  Scalar(_, node) {
    delete node.format
    delete node.type
  }
}
```
and then apply it:
```
echo '{"hello":"world"}' | npx yaml --visit ./strip-styles.mjs
```
for this output:
```yaml
hello: world
```

danielbayley:
@eemeli Nice! I had actually already worked up a PR for a CLI, but got sidelined… This is working well, though missing the option to output pretty-printed JSON… Hence my PR in #559.

eemeli:
> Also to reiterate what I mentioned in [#479 (comment)](https://github.com/eemeli/yaml/issues/479#issuecomment-1690237880):
> 
> > it should include aliases to seamlessly replace `yaml2json` and `json2yaml` that [come with](https://github.com/jeremyfa/yaml.js#command-line-tools) _[yamljs](https://github.com/jeremyfa/yaml.js#readme)_ [here](https://github.com/jeremyfa/yaml.js/blob/develop/package.json#L22-L25), making replacing that library with this (better one) easier…
> 
> which I think would help adoption…

Please submit this as a separate issue.

PR #542, opened 2024-05-21
Update public-api.ts

bluelovers:


eemeli:
Closing, as there's no explanation for the change, and no added functionality that wouldn't be fulfilled by an explicit `as T` cast in the calling context.

bluelovers:
`(parse() as T).xxx` vs `parse<T>().xxx`

PR #559, opened 2024-06-30
feat: Add --indent option to CLI tool

danielbayley:
Add a `--indent` option, to be used with `--json` to output 2 space indentation.

See also #479 and #523.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,7 +24,7 @@ Usage:
 Options:
   --help, -h    Show this message.
   --json, -j    Output JSON.
-  --pretty, -p  Output pretty-printed JSON.
+  --indent 2    Output pretty-printed data, indented by the given number of spaces.
 
 Additional options for bare "yaml" command:
   --doc, -d     Output pretty-printed JS Document objects.
@@ -56,8 +56,8 @@ export async function cli(
       options: {
         doc: { type: 'boolean', short: 'd' },
         help: { type: 'boolean', short: 'h' },
+        indent: { type: 'string', short: 'i' },
         json: { type: 'boolean', short: 'j' },
-        pretty: { type: 'boolean', short: 'p' },
         single: { type: 'boolean', short: '1' },
         strict: { type: 'boolean', short: 's' },
         visit: { type: 'string', short: 'v' },
@@ -72,6 +72,8 @@ export async function cli(
     positionals: [mode],
     values: opt
   } = args
+
+  let indent = Number(opt.indent)
 
   stdin.setEncoding('utf-8')
 
@@ -94,7 +96,7 @@ export async function cli(
       })
       stdin.on('end', () => {
         for (const tok of lexer.lex('', false)) add(tok)
-        if (opt.json) console.log(JSON.stringify(data))
+        if (opt.json) console.log(JSON.stringify(data, null, indent))
         done()
       })
       break
@@ -112,7 +114,7 @@ export async function cli(
       })
       stdin.on('end', () => {
         for (const tok of parser.parse('', false)) add(tok)
-        if (opt.json) console.log(JSON.stringify(data))
+        if (opt.json) console.log(JSON.stringify(data, null, indent))
         done()
       })
       break
@@ -161,7 +163,8 @@ export async function cli(
         } else {
           if (reqDocEnd) console.log('...')
           try {
-            const str = String(doc)
+            indent ||= 2
+            const str = doc.toString({ indent })
             console.log(str.endsWith('\n') ? str.slice(0, -1) : str)
           } catch (error) {
             done(error as Error)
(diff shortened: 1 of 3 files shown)


eemeli (changes requested):
I have two concerns with this change:

1. Could you clarify what's the use case of pretty-printed JSON output? For single-shot manual use, wouldn't the same be served by piping the output to a separate JSON prettifier like `jq`?

2. Naming the option as "pretty" and only affecting the JSON output seems rather misleading. Either the option needs to make sense independently of `--json`, or it needs to be named more specifically.

eemeli:
> > 1. Could you clarify what's the use case of pretty-printed JSON output? For single-shot manual use, wouldn't the same be served by piping the output to a separate JSON prettifier like `jq`?
> 
> _Non_ single-shot manual use… Like in `package.json` ([`package.yaml`](https://github.com/pnpm/pnpm/issues/1100)) `scripts`.

Eh, not sure that it really makes sense to prettify output that's meant for machine consumption, as it risks creating JSON files that will be presumed to be the source of truth, rather than intermediate build assets. But I'll grant you that there's probably a valid use case somewhere here.

> > 2. Naming the option as "pretty" and only affecting the JSON output seems rather misleading. Either the option needs to make sense independently of `--json`, or it needs to be named more specifically.
> 
> I had a similar thought, but there is no such thing as `--pretty` YAML, it already has to be indented…
> 
> I think the ideal API would be an optional parameter, maybe to `--json`, e.g `--json 2` or something, but that isn’t really a feature of `util.parseArgs`. How about `--indent`? What other name do you think would be better?

`--indent` works, esp. as it's already an [option](https://eemeli.org/yaml/#tostring-options) for YAML formatting. Switch this PR to that and make it apply to YAML output as well?

danielbayley:
> it risks creating JSON files that will be presumed to be the source of truth

Well the YAML files are the source… But a _lot_ of tools only work from JSON. Harder to debug if it _has_ to be minified also…

> `--indent` works, esp. as it's already an [option](https://eemeli.org/yaml/#tostring-options) for YAML formatting. Switch this PR to that and make it apply to YAML output as well?

Good idea, will look into it…

src/cli.ts
```diff
@@ -57,11 +57,11 @@ export async function cli(
         doc: { type: 'boolean', short: 'd' },
         help: { type: 'boolean', short: 'h' },
         json: { type: 'boolean', short: 'j' },
-        pretty: { type: 'boolean', short: 'p' },
         single: { type: 'boolean', short: '1' },
         strict: { type: 'boolean', short: 's' },
         visit: { type: 'string', short: 'v' },
-        yaml: { type: 'string', default: '1.2' }
+        yaml: { type: 'string', default: '1.2' },
+        indent: { type: 'string', short: 'i' }
```
eemeli, line 64: Should keep these in alphabetic order.

src/cli.ts
```diff
@@ -24,6 +24,7 @@ Usage:
 Options:
   --help, -h    Show this message.
   --json, -j    Output JSON.
+  --pretty, -p  Output pretty-printed JSON.
```
eemeli, line 27: Needs an update.

src/cli.ts
```diff
@@ -188,7 +190,7 @@ export async function cli(
           )
         }
         if (mode !== 'valid' && opt.json) {
-          console.log(JSON.stringify(opt.single ? data[0] : data))
+          console.log(JSON.stringify(opt.single ? data[0] : data, null, Number(opt.indent)))
```
eemeli, line 193: Note also the other places where output is being generated: lines 97, 115, and 164 will need to be updated. It'd probably be best to determine a single `jsonIndent` value somewhere above, and use that for the `JSON.stringify()` calls, and handle the yaml Document's `toString{ indent })` value separately.

tests/cli.ts
eemeli, line 1: The lexer, CST, and YAML document outputs will also need `--indent` tests.

eemeli (changes requested):
Overall the API looks good, see inline for implementation changes.

tests/cli.ts
danielbayley: All passing, except something weird going on with the CST unit test… Whether I hard code it as a string, or do `JSON.stringify` also in the test. Any ideas?

tests/cli.ts
```diff
@@ -114,6 +114,94 @@ const skip = Number(major) < 20
...
+      )
+      ok(
+        'Lexer',
+        'hello: world',
+        ['lex', '--json', '--indent', '2'],
+        ['[\n  "\\u0002",\n  "\\u001f",\n  "hello",\n  ":",\n  " ",\n  "\\u001f",\n  "world"\n]']
+      )
+      ok(
+        'CST parser',
+        'hello: world',
```
eemeli, line 150: The expected output includes a trailing newline, which is missing from here.

```suggestion
        'hello: world\n',
```

eemeli (approved):
Thank you, this looks good! Sorry for the delay in getting to this. There was a missing newline from the test source, which I added to get the test to pass.

danielbayley:
> Thank you, this looks good! Sorry for the delay in getting to this. There was a missing newline from the test source, which I added to get the test to pass.

No worries, life gets in the way… I know! Ah cool, was that at the end?

PR #587, opened 2024-11-12
Update schema.ts

vra5107:
CodeQL is complaining about missing js regex anchors.

`Misleading operator precedence. The subexpression '^true' is anchored at the beginning, but the other parts of this regular expression are not
Misleading operator precedence. The subexpression 'false$' is anchored at the end, but the other parts of this regular expression are not`

https://codeql.github.com/codeql-query-help/javascript/js-regex-missing-regexp-anchor/

### DIFF
--- a/src/schema/json/schema.ts
+++ b/src/schema/json/schema.ts
@@ -30,7 +30,7 @@ const jsonScalars: ScalarTag[] = [
     identify: value => typeof value === 'boolean',
     default: true,
     tag: 'tag:yaml.org,2002:bool',
-    test: /^true|false$/,
+    test: /^true$|^false$/,
     resolve: str => str === 'true',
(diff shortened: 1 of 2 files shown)


eemeli:
Huh, that is indeed a bug. Should probably add a test or two for this as well, probably around [here](https://github.com/eemeli/yaml/blob/2e85b914b918b13e3985c58b1ce735d2c3f5a813/tests/doc/types.ts#L268). Would you be up to doing so?

vra5107:
@eemeli I have added a test. Did run it locally and it does pass. It does return an error for something like `ffalse` but reports no errors for something like `Truea`. If you can kindly merge and release a new version, I will pull the latest version. Security is a mania these days.

eemeli (approved):

PR #612, opened 2025-03-28
feat: Build cached anchor/alias node list for faster alias resolving

PolyMeilex:
Currently, to resolve an alias, we have to visit the whole document tree until the alias node is found. The lower in the file we put an alias, the slower the resolution is.

This is an attempt to speed this process significantly by:
1. Putting all nodes in a list and iterating over it, this is significantly faster than descending the tree via visitor every time
2. If we are putting nodes into a list we might as well filter them to only the ones that are needed for alias resolution, so only aliases and anchors are cached

Not sure if this is the right approach, but it helped me reduce the load time of my dataset, which is not very alias-heavy, from 10s to 4s.

### DIFF
--- a/src/nodes/Alias.ts
+++ b/src/nodes/Alias.ts
@@ -42,23 +42,21 @@ export class Alias extends NodeBase {
     doc: Document,
     ctx?: ToJSContext
   ): Scalar | YAMLMap | YAMLSeq | undefined {
-    let found: Scalar | YAMLMap | YAMLSeq | undefined = undefined
-
-    if (!ctx) {
+    let nodes: Node[]
+    if (ctx?.aliasResolveCache) {
+      nodes = ctx.aliasResolveCache
+    } else {
+      nodes = []
       visit(doc, {
         Node: (_key: unknown, node: Node) => {
-          if (node === this) return visit.BREAK
-          if (node.anchor === this.source) found = node
+          if (isAlias(node) || hasAnchor(node)) nodes.push(node)
         }
       })
-      return found
+      if (ctx) ctx.aliasResolveCache = nodes
     }
 
-    if (!ctx.aliasResolveCache) {
-      ctx.aliasResolveCache = buildCache(doc)
-    }
-
-    for (const node of ctx.aliasResolveCache) {
+    let found: Scalar | YAMLMap | YAMLSeq | undefined = undefined
+    for (const node of nodes) {
       if (node === this) break
       if (node.anchor === this.source) found = node
     }
@@ -138,16 +136,4 @@ function getAliasCount(
     return Math.max(kc, vc)
   }
   return 1
-}
-
-function buildCache(
-  doc: Document<Node, boolean>
-): (Scalar | YAMLMap | YAMLSeq | Alias)[] {
-  const found: (Scalar | YAMLMap | YAMLSeq | Alias)[] = []
-  visit(doc, {
-    Node: (_key: unknown, node: Node) => {
-      if (isAlias(node) || hasAnchor(node)) found.push(node)
-    }
-  })
-  return found
 }
--- a/src/nodes/Alias.ts
+++ b/src/nodes/Alias.ts
@@ -3,7 +3,7 @@ import type { Document } from '../doc/Document.ts'
 import type { FlowScalar } from '../parse/cst.ts'
 import type { StringifyContext } from '../stringify/stringify.ts'
 import { visit } from '../visit.ts'
-import { ALIAS, isAlias, isCollection, isPair } from './identity.ts'
+import { ALIAS, hasAnchor, isAlias, isCollection, isPair } from './identity.ts'
 import type { Node, Range } from './Node.ts'
 import { NodeBase } from './Node.ts'
 import type { Scalar } from './Scalar.ts'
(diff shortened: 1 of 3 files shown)


src/nodes/Alias.ts
```diff
@@ -52,7 +52,17 @@ export class Alias extends NodeBase {
...
+
+    let source: Scalar | YAMLMap | YAMLSeq | undefined = undefined
+    if (ctx.anchorAndAliasNodes) {
+      for (const node of ctx.anchorAndAliasNodes) {
+        if (node === this) break
+        if (node.anchor === this.source) source = node
+      }
+    } else {
+      source = this.resolve(doc)
+    }
```
eemeli, line 64: This duplication of logic should be in the `resolve()` method, rather than here. Add a second optional `ctx` argument to the method?

src/nodes/toJS.ts
```diff
@@ -10,6 +14,8 @@ export interface AnchorData {
 
 export interface ToJSContext {
   anchors: Map<Node, AnchorData>
+  /** Cached anchor and allias nodes in the order they occurred in the document */
+  anchorAndAliasNodes?: (Alias | Scalar | YAMLMap | YAMLSeq)[]
```
eemeli, line 18: Why is this optional? It might also be better to name it to correspond to its use, i.e. something like `aliasResolveCache` or similar. Not sure though.

src/nodes/toJS.ts
```diff
@@ -33,13 +39,24 @@ export function toJS(value: any, arg: string | null, ctx?: ToJSContext): any {
...
     // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    if (!ctx || !hasAnchor(value)) return value.toJSON(arg, ctx)
+    if (!ctx) return value.toJSON(arg, ctx)
+
+    if (!hasAnchor(value)) {
+      if (ctx.anchorAndAliasNodes && isAlias(value))
+        ctx.anchorAndAliasNodes.push(value)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      return value.toJSON(arg, ctx)
+    }
```
eemeli, line 49: Could this also be done within the Alias `resolve()`?

eemeli (changes requested):
I can see how this might be helpful. I'd much prefer for most of the work to be done in the Alias `resolve()` method, though. See inline for further comments.

Would you be able to share some YAML file with which the performance benefits of the change become measurable?

src/nodes/toJS.ts
PolyMeilex: Renamed, at first I made it optional so that the change is non-api braking, but in the new version I use it to figure out if cache was already built or not. If it is undefined cache needs to be built.

src/nodes/Alias.ts
PolyMeilex: Moved to resolve

src/nodes/toJS.ts
PolyMeilex: Yep, moved

PolyMeilex:
> Would you be able to share some YAML file with which the performance benefits of the change become measurable?

Had to anonymize it, so it looks kinda goofy, but here is one of the worst offenders in my data set: https://pastebin.com/nPB4hLyT

src/nodes/Alias.ts
```diff
@@ -38,21 +38,38 @@ export class Alias extends NodeBase {
...
+      ctx.aliasResolveCache = buildCache(doc)
+    }
+
+    for (const node of ctx.aliasResolveCache) {
+      if (node === this) break
+      if (node.anchor === this.source) found = node
+    }
+
     return found
   }
```
eemeli, line 67: We can presume that calling `resolve()` without a context is really rare, so this can be simpler, something like:
```ts
  resolve(
    doc: Document,
    ctx?: ToJSContext
  ): Scalar | YAMLMap | YAMLSeq | undefined {
    let nodes: Node[]
    if (ctx?.aliasResolveCache) {
      nodes = ctx.aliasResolveCache
    } else {
      nodes = []
      visit(doc, {
        Node: (_key: unknown, node: Node) => {
          if (isAlias(node) || node.anchor) nodes.push(node)
        }
      })
      if (ctx) ctx.aliasResolveCache = nodes
    }

    let found: Scalar | YAMLMap | YAMLSeq | undefined = undefined
    for (const node of nodes) {
      if (node === this) break
      if (node.anchor === this.source) found = node
    }
    return found
  }
```

src/nodes/toJS.ts
```diff
@@ -10,6 +14,8 @@ export interface AnchorData {
 
 export interface ToJSContext {
   anchors: Map<Node, AnchorData>
+  /** Cached anchor and allias nodes in the order they occurred in the document */
+  aliasResolveCache?: (Alias | Scalar | YAMLMap | YAMLSeq)[]
```
eemeli, line 18: The typing gets restricted from `Node` by the `node.anchor` check, so this doesn't need to list all the sub-types.

```suggestion
  /** Cached anchor and alias nodes in the order they occur in the document */
  aliasResolveCache?: Node[]
```

eemeli (changes requested):
Getting there, I think. And thank you for the sample data, I'll need to find a bit of time to test how much this improves the performance on it.

PolyMeilex:
>  I'll need to find a bit of time to test how much this improves the performance on it.

Not a proper scientific benchmark, but with the example file this is the difference on my setup:

`before`
```
parse: 141.241ms
toJs: 1.684s
```
`after`
```
parse: 145.326ms
toJs: 26.266ms
```

src/nodes/Alias.ts
PolyMeilex: Sure, looks a lot better. Done

src/nodes/toJS.ts
PolyMeilex: Done

eemeli: Note the typo fixes in the comment, those still remain.

PolyMeilex: Oh, my bad, fixed

PolyMeilex:
We could also build a map where a key would be the anchor's name and the value would be the node with all its child aliases already resolved. Then if a duplicated anchor name is found it would simply replace the previous one, but I'm assuming that there is some reason why that's not how it's done that I'm not aware of.

eemeli (approved):
Thank you @PolyMeilex for your contribution!

With you test file, this does provide something like a 12x improvement when I test it locally, though processing it does require adjusting `maxAliasCount` to avoid detecting the input as a resource exhaustion attack.

> We could also build an anchors map [...] but I'm assuming that there is some reason why that's not how it's done already, that I'm not aware of.

That's effectively `ctx.anchors`, but it's not as simple as one might like. The YAML spec allows for fun structures like this:

```yaml
&foo
key:
- *foo
- &foo 42
- *foo
```

Here, the resolution of the top-level map depends on resolving all of its contents, and that includes resolving the first `*foo` as a self-referential reference to the map itself, as well as resolving the second `*foo` as the scalar `42`, thanks to the anchor re-use.

Getting that to work right was a bit tricky.

PolyMeilex:
> That's effectively `ctx.anchors`, but it's not as simple as one might like. The YAML spec allows for fun structures like this:
> 
> ```yaml
> &foo
> key:
> - *foo
> - &foo 42
> - *foo
> ```
> 
> Here, the resolution of the top-level map depends on resolving all of its contents, and that includes resolving the first `*foo` as a circular reference to the map itself, as well as resolving the second `*foo` as the scalar `42`, thanks to the anchor re-use.
> 
> Getting that to work right was a bit tricky.

Oh, yeah that's fun, would rather pay the price of wasteful iteration then try to implement this 😅

Thank you for being open to this! I will try to find more low hanging fruit so hopefully we can keep generators, while still making the package  usable for my usecase 🚀

PR #613, opened 2025-03-29
feat: add indentBeforeInlineComment ToStringOption

AlexisColes:
Adding a new ToStringOption, indentBeforeInlineComment, to provide more flexibility on yaml output styling.

We already have linting conventions in place that conflict with the yaml being output by this library.  It would be nice to have the flexibility get the output to match the existing conventions.

We need to be able to specify that block comments have 2 not 1 space between the value and the start of the comment

```
key: value  # my comment
```
Naming is hard, would appreciate any better suggestions you may have for this config option !

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -323,6 +323,13 @@ export type ToStringOptions = {
   indent?: number
 
   /**
+   * The number of spaces to use when indenting an inline comment.
+   *
+   * Default: `1`
+   */
+  indentBeforeInlineComment?: number
+
+  /**
    * Whether block sequences should be indented.
    *
    * Default: `true`
--- a/src/stringify/stringify.ts
+++ b/src/stringify/stringify.ts
@@ -26,6 +26,7 @@ export type StringifyContext = {
   indentStep: string
   indentAtStart?: number
   inFlow: boolean | null
+  inlineIndentBeforeComment?: number,
   inStringifyKey?: boolean
   flowCollectionPadding: string
   options: Readonly<
@@ -50,6 +51,7 @@ export function createStringifyContext(
       falseStr: 'false',
       flowCollectionPadding: true,
       indentSeq: true,
+      inlineIndentBeforeComment: 1,
       lineWidth: 80,
       minContentWidth: 20,
       nullStr: 'null',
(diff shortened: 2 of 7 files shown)


eemeli (changes requested):
Tbh, this seems somewhat niche. YAML allows for a lot of optionality with spaces and indents, and some of that is indeed configurable in the `yaml` output, but I would prefer not to add much more.

If there is a need for more customizability of the output, it might be more interesting to solve that by providing an intermediate output format that could then be processed as proposed here. That would also allow for e.g. comment alignment, which is also not currently supported.

Or is this specific need something that is shared more widely?

AlexisColes:
I do take your point, where does all this configuration end!  

I would imagine we are not the only ones that have run into this problem.  It was not us that decided that 2 spaces was the golden number in this situation.  We are just using the default rules from [yamllint](https://github.com/adrienverge/yamllint/blob/bce26d68a881551bb448114fbbf7fc472e3feb9e/yamllint/rules/comments.py#L83) via the [chart-testing-action](the https://github.com/helm/chart-testing-action).  These are widely used tools that have decided that 2 spaces is the way to go for these things.  

We did try to change the configuration for yamllint to change min-spaces-from-content to 1, but it was not being respected when used via the github action for some reason.  We could probably get to the bottom of this but in this case it seemed a better solution to try and get the conflict fixed on this side.

TBH the ideal solution would be if we could have an option that didn't change the formatting of the existing document at all.  All we are trying to do is update one value in an existing yaml file. 

But that being said if we ever did want to add an inline comment would be good if we could do it in a way that played nice with these mainstream tools.

eemeli:
> TBH the ideal solution would be if we could have an option that didn't change the formatting of the existing document at all. All we are trying to do is update one value in an existing yaml file.

If that's what you're looking to do, then the better approach would be to mutate and stringify the CST representation of the document. So something like this:

```js
import { Composer, CST, Parser } from 'yaml'

const src = 'key: value  # my comment'
const cst = Array.from(new Parser().parse(src));
const [doc] = Array.from(new YAML.Composer({ keepSourceTokens: true }).compose(cst));

// modify the value in the CST
const node = doc.get('key', true);
node.srcToken.source = 'new value';

const res = cst.map(CST.stringify).join('');
// 'key: new value  # my comment'
```

Note that the CST is not validated in any way during serialization, so setting the `source` to a value that isn't valid YAML for that position in the document will break things.

AlexisColes:
I have just given this a go, this is perfect for what we need.  TYSM!! :raised_hands: 

Lets close this one off, your solution is way better!!

PR #615, opened 2025-04-14
Respect Node.type === 'PLAIN' for numeric-like strings

edriouk:
This pull request fixes bug #616
The following changes are made:
A check is added if node type is  'PLAIN' in plainString() function ensuring it is respected.
Unit tests are added to check if strings looking numeric (hexadecimal, decimal and float) and boolean values are output as plain strings if node type is  'PLAIN' is specified. The tests also ensure that default behavior has not changed. 
Two regression tests in tests/doc/types.ts are corrected to check new behavior.
All other tests have passed without modifications.


eemeli:
See parent issue for alternative solution.

PR #617, opened 2025-04-16
docs: include cli example

JAdshead:
Adding basic example of CLI usage.

Reading from STDIN is not pattern I have used for node based CLI tools and took me a second to process. Example might help someone else.

Feel free to decline if its not aligned.

### DIFF
--- a/docs/09_cli.md
+++ b/docs/09_cli.md
@@ -1,12 +1,12 @@
 # Command-line Tool
 
-Available as `npx yaml` or `npm exec yaml`:
-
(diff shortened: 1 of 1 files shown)


eemeli (approved):
Sorry for taking so long to get at this, but this is fine. I moved the block one paragraph earlier so that it's more clearly separate from the cli-help block.

JAdshead:
No worries. I appreciate the effort it takes being a maintainer so thank you.

Issue #619, opened 2025-05-16
Strings like "10:00" are serialized to mean something different between YAML 1.1 and 1.2 readers


jwhitaker-gridcog:
**Describe the bug**
If I run
```js
stringify({key: "10:00"})
```

then the result is

```yaml
key: 10:00
```
.

That yaml is unsafe - if you read it with a 1.2 yaml parser it means

```js
{"key": "10:00"}
```

but if you read it with a 1.1 yaml parser it means

```js
{"key": 600}
```

.

**Expected behaviour**
Given `yaml` seems to take consistent behaviour between 1.1 and 1.2 very seriously, it would be very nice if this was consistently serialized as 

```yaml
key: "10:00"
```

.

**Versions (please complete the following information):**

- Environment: Nodejs 20
- `yaml`: 2.7.1

Keywords for googlers: sexagesimal 
Ref: https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell#sexagesimal-numbers

eemeli:
If this is a concern for you, you should use the `compat` [schema option](https://eemeli.org/yaml/#schema-options) when serialising:
```js
stringify({key: "10:00"})
// 'key: 10:00\n'

stringify({ key: '10:00' }, { compat: 'yaml-1.1' })
// 'key: "10:00"\n'
```

jwhitaker-gridcog:
thanks heaps! i had missed this in the doco.

PR #621, opened 2025-05-18
docs: add a warning for stringify vs. pyyaml

jwhitaker-gridcog:
Heya, here's a proposal for a doco change c.f. #619 . I had read "full support for 1.1 and 1.2" in the project intro so had incorrectly assumed that the default settings would also meet those requirements, which isn't the case. imo it could be worth adding a warning at top-level `stringify` doco to try and make this clear?
Cheers

### DIFF
--- a/docs/02_parse_stringify.md
+++ b/docs/02_parse_stringify.md
@@ -69,6 +69,8 @@ YAML.stringify({ number: 3, plain: 'string', block: 'two\nlines\n' })
 
 `value` can be of any type. The returned string will always include `\n` as the last character, as is expected of YAML documents. If defined, the `replacer` array or function follows the [JSON implementation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter). See [Options](#options) for more information on the last argument, an optional configuration object. For JSON compatibility, using a number or a string as the `options` value will set the `indent` option accordingly.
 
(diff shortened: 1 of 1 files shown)


docs/02_parse_stringify.md
```diff
@@ -69,6 +69,8 @@ YAML.stringify({ number: 3, plain: 'string', block: 'two\nlines\n' })
 
 `value` can be of any type. The returned string will always include `\n` as the last character, as is expected of YAML documents. If defined, the `replacer` array or function follows the [JSON implementation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter). See [Options](#options) for more information on the last argument, an optional configuration object. For JSON compatibility, using a number or a string as the `options` value will set the `indent` option accordingly.
 
+Note that the default output of `stringify` is not safe to pass to YAML 1.1 consumers, e.g. PyYAML. You should set [`compat: true`](#schema-options) in `options` if you are targeting these consumers.
```
perlpunk, line 72: I wouldn't specifically mention a certain library here.
It's true that PyYAML itself does not support 1.2 at this point in time, but there is a library for that which you can use on top of PyYAML, and at some point it might be merged upstream.
(just IMHO as a regular pyyaml contributor)

jwhitaker-gridcog: fwiw I called out pyyaml here because (anecdotally) it's a very common consumer, and (anecdotally) its users often aren't even aware that YAML 1.1 and 1.2 exist, much less the differences in parsing behaviour between them and which one pyyaml uses. As such I suspected the more correct "only works for yaml 1.2 readers" wouldn't be sufficient to warn many people who would be tripped up by this.

PR #632, opened 2025-07-11
feat: yaml.parse has generic type for return type

Uzlopak:
I am migrating a repository from js-yaml to yaml. I saw that yaml has typescript typings, but the parse operation has no generic type for the return type. 

So here I propose to add a return Type.

before:

```typescript
const manifest = yaml.parse(file) as Manifest;
```

```typescript
const manifest = yaml.parse<Manifest>(file);
```

By doing so, I also get feedbeck by typescript, that manifest is potentially null, and I have to verify that I have not null as a result. 

Just because of this, I will need to check if js-yaml and yaml differ on parsing errors. I guess js-yaml, throws a YAMLException but yaml just returns null? Anyway. I think this is a usefull feature.

### DIFF
--- a/src/public-api.ts
+++ b/src/public-api.ts
@@ -131,23 +131,23 @@ export function parseDocument<
  *   document, so Maps become objects, Sequences arrays, and scalars result in
  *   nulls, booleans, numbers and strings.
  */
-export function parse(
+export function parse<R = any>(
   src: string,
   options?: ParseOptions & DocumentOptions & SchemaOptions & ToJSOptions
-): any
-export function parse(
+): R | null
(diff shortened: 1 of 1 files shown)


eemeli:
I won't be accepting this proposed change.

The `YAML.parse()` function is meant to behave as closely as possible to `JSON.parse()`, and in this aspect it means typing it the same way as TypeScript does in `lib/lib.es5.d.ts`, i.e. returning `any`. It really ought to be `unknown` tbh, but sticking to precent is more important here.

When calling `parse()` with bad input, an error will be thrown.

Issue #630, opened 2025-06-30
For an empty multi line input like |- is getting replace with ''


yadavjay8887:
**Describe the bug**
For an empty multi line input like |- is getting replace with ''

**To Reproduce**
Steps to reproduce the behaviour.

When parsing a YAML document containing an empty block scalar using `|-` (block literal).
Then serializing it back with `.toString()`, the scalar is converted to a plain quoted string (`''`).
Losing its original formatting.

**Input YAML
input: |-

Output YAML
input: ''**

**Expected behaviour**

Input YAML

input: |-

Output YAML

input: |-

**Versions (please complete the following information):**

- Environment: 
- Node.js 22.14.0
- `yaml`: 2.6.1

**Additional context**

eemeli:
Please share the actual JS code you're using to get this result; if you're using `parse()` rather than `parseDocument()`, the presentation details are discarded.

yadavjay8887:
I am using electron js with node and angular 
Below is the typescript function

`async readYamlFromPath(filePath: string) {
  let fileData = await this.ipcRenderer.sendEvent('readFile', filePath);
  let testDoc: any = parseDocument(fileData);
  console.log(testDoc.toString());
}
`

Issue is when there is no input string and only |- is present as shown in case of Input1 and Output1
If it has string values then it works fine as shown in case of Input2 and Output2

Input 1
parent:
  child: |-

Output 1
parent:
  child: ""

Input 2
parent:
  child: |-
    test1
    test2

Output 2
parent:
  child: |-
    test1
    test2

eemeli:
Ah, now I see it; this is the issue you're seeing, and it is indeed a bit surprising:
```js
parseDocument('|-\n').toString() === '""\n'
```
There's no reason to discard the preferred `BLOCK_LITERAL` type there. If you're interested in providing a fix yourself, this is probably the place to work on: https://github.com/eemeli/yaml/blob/cad823ebd805aae2e9b743a5fb89f07a06cf505d/src/stringify/stringifyString.ts#L187-L191

PR #634, opened 2025-07-25
preserve empty block literals

Shruti157:
# Fix: Preserve empty block literals (`|-`) instead of converting to empty strings

## Description

This PR fixes the issue where empty block literals using the `|-` chomping indicator are incorrectly stringified as quoted empty strings (`''`) instead of retaining their original block literal formatting.

Currently, this input:

```yaml
input: |-   
```

is stringified as:

```yaml
input: ''
```
---

## Related Issue

Fixes [#630](https://github.com/eemeli/yaml/issues/630)

---

## What’s the Fix?

The following condition in `blockString()` was:

```ts
if (!blockQuote || /\n[\t ]+$/.test(value) || /^\s*$/.test(value)) {
  return quotedString(value, ctx)
}

if (!value) return literal ? '|\n' : '>\n'
```

This logic caused empty or whitespace-only block scalars to fallback to quoted strings (`''`), even when the user clearly wrote a `|-` block scalar.

### Updated To:

```ts
if (!blockQuote || /\n[\t ]+$/.test(value)) {
  return quotedString(value, ctx)
}

if (!value) return literal ? '|-\n' : '>\n'
```

### Key Change:
- **Removed check** for `^\s*$` which incorrectly matched valid empty block literals.
- **Changed fallback** to return `|-` instead of `|` to match typical YAML idioms and avoid unexpected trailing newlines.

---

## Result

With this fix, the following roundtrip works as expected:

```yaml
input: |-   
```

After parse → stringify:

```yaml
input: |-   
```

✅ **Block literal preserved**

---

## Additional Notes

- Tested against YAML samples including PEM-style certificates and template files.

### DIFF
--- a/tests/doc/stringify.ts
+++ b/tests/doc/stringify.ts
@@ -330,6 +330,36 @@ z:
     expect(String(doc)).toBe('? >+ #comment\n  foo\n\n: bar\n')
   })
 
+  test('Keep pipe when empty block literal is given', () => {
+    const doc = YAML.parseDocument('|-')
+    expect(String(doc)).toBe('|\n\n')
+  })
+
+  test('Keep pipe when pipe is provided', () => {
+    const doc = YAML.parseDocument('|')
+    expect(String(doc)).toBe('|\n\n')
+  })
+
+  test('Keep pipe when empty block literal is given yaml format', () => {
+    const src = source`
+      log4j_2: |-
+    `;
+    const expected = source`
+      log4j_2: |
+
+    `;
+    const doc = YAML.parseDocument(src)
+    expect(String(doc)).toBe(expected)
+  })
+
+  test('Keep pipe when pipe is provided in yaml format', () => {
+    const src = source`
+      log4j_2: |
+    `
+    const doc = YAML.parseDocument(src)
+    expect(String(doc)).toBe(src+'\n')
+  })
+
   test('Document as key', () => {
     const doc = new YAML.Document({ a: 1 })
     doc.add(new YAML.Document({ b: 2, c: 3 }))
--- a/tests/doc/stringify.ts
+++ b/tests/doc/stringify.ts
@@ -186,6 +186,16 @@ blah blah\n`)
             .map(line => line.length)
         ).toMatchObject([4, 20, 20, 20, 20, 10, 0])
       })
+
+      test('Keep pipe when empty block literal is given', () => {
+        const doc = YAML.parseDocument('|-')
+        expect(String(doc)).toBe('|\n\n')
+      })
+
+      test('Keep pipe when pipe is provided', () => {
+        const doc = YAML.parseDocument('|')
+        expect(String(doc)).toBe('|\n\n')
+      })
     })
   })
 }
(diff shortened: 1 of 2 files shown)


Shruti157:
Yes, from a parsing perspective both | and |- parse to the same empty string.
However, this change ensures that if a user writes input: |-, that exact formatting gets preserved when stringified back. Currently |- becomes |, which changes their original formatting.
The - indicator shows explicit intent to strip trailing newlines. Even for empty values, preserving this keeps the original meaning intact.
This makes behavior consistent with non-empty block literals where chomping indicators are correctly preserved.

Would you suggest a different approach for handling this type of empty block literal case?

eemeli:
I continue to have a really strong sense that I'm interacting here with an LLM here, and I'd really rather not. In fact, this PR inspired me to add this to the [contribution guide](https://github.com/eemeli/yaml/blob/main/docs/CONTRIBUTING.md#contributing-code):
> If you use an LLM when writing pull requests, this must be explicitly declared in the pull request. If there is indication of undeclared LLM assistance, the pull request will be declined.

With that in mind, please clarify if you're using an LLM to write your comments. And going forward, I'd be much more willing to consider your ideas if you represent them directly (even if it's in bad English, I honestly don't mind) rather than if I have to presume they've been generated by an LLM.

rakumar01:
We're kind of block on this issue and not very versed with the library, so we have been quite detailed in explaining our requirements in the PR, which might come across as Llmish. We're looking for suggestions on how to move forward. Going forward, we'll keep our explanations more concise.

rakumar01:
We have spent the last three days trying to understand the library to provide a fix and validated test case scenarios for this PR. If this PR is being rejected mainly due to the tone or what seems like LLM-ish verbosity, please do let us know. It seems that the style of the explanation is getting more attention than the actual problem, especially when it is blocking us. We are blocked due to an issue in the library which we had already explained and want to unblock ourselves with the provided fix. If the fix itself needs work, we are happy to discuss and improve it. If it's mainly about the writing and the tone, let us know so we can revise and resubmit the PR with a more natural tone to avoid rejection over verbosity.

rakumar01:
Thanks for the insights and clarification on the Document level API limitations. We will  revise our code to remove the - indicator from the condition while keeping the block literal format. We will also explore the CST API for finer control.

Shruti157:
we have removed the '-' indicator from empty block literal as discussed. The block literal format is still retained. Please review.

Shruti157:
Hi @eemeli,

I have added test cases to verify the behavior with my changes.

tests/doc/stringify.ts
```diff
@@ -330,6 +330,36 @@ z:
...
 
+  test('Keep pipe when empty block literal is given', () => {
+    const doc = YAML.parseDocument('|-')
+    expect(String(doc)).toBe('|\n\n')
+  })
+
+  test('Keep pipe when pipe is provided', () => {
+    const doc = YAML.parseDocument('|')
+    expect(String(doc)).toBe('|\n\n')
+  })
```
eemeli, line 341: These are testing scalar string serialization, so don't belong in this block that's testing map serializations. Move these to somewhere near the end of the block starting on line 107.

tests/doc/stringify.ts
```diff
@@ -330,6 +330,36 @@ z:
...
+    expect(String(doc)).toBe(expected)
+  })
+
+  test('Keep pipe when pipe is provided in yaml format', () => {
+    const src = source`
+      log4j_2: |
+    `
+    const doc = YAML.parseDocument(src)
+    expect(String(doc)).toBe(src+'\n')
+  })
```
eemeli, line 361: "yaml format" doesn't really mean anything in this context, and these don't add any value beyond the first two tests; drop these.

eemeli (changes requested):

Shruti157:
Changes have been incorporated. Please review.

eemeli (approved):
This is fine. Not 100% happy about the trailing `\n\n` for top-level block scalars, but that's not really that significant.

PR #642, opened 2025-11-21
fix: Ensure single newline after key with empty value in stringify output

### LINKED ISSUE #474: Extra newline is added after empty values
**Describe the bug**
The library adds an additional newline after an empty value and a newline

**To Reproduce**
``` js
import { parseDocument } from 'yaml';
import { createPatch } from 'diff';

const yaml = `
map:
  item:

anothermap:
  anotheritem:

`;

const document = parseDocument(yaml);
const result = document.toString();

const patch = createPatch('output', yaml, result);

console.log(patch);
```

Output:
``` patch
Index: output
===================================================================
--- output
+++ output
@@ -1,7 +1,8 @@
-
 map:
   item:

+
 anothermap:
   anotheritem:

+
```

**Expected behaviour**
Expected output:
``` patch
Index: output
===================================================================
--- output
+++ output
```

- Environment: Node v16.16.0
- `yaml`: 2.2.2

Neutronlul:
DISCLAIMER: I based this change off of [code generated by an LLM](https://github.com/Neutronlul/yaml/pull/1).

This pull request attempts to fix #474.

Using test code taken from #474:

```javascript
import { parseDocument } from 'yaml';
import { createPatch } from 'diff';

const yaml = `
map:
  item:

anothermap:
  anotheritem:

`;

const document = parseDocument(yaml);
const result = document.toString();

const patch = createPatch('output', yaml, result);

console.log(patch);
```

Before:

```diff
Index: output
===================================================================
--- output
+++ output
@@ -1,7 +1,8 @@
-
 map:
   item:

+
 anothermap:
   anotheritem:

+
```

After:

```diff
Index: output
===================================================================
--- output
+++ output
@@ -1,5 +1,4 @@
-
 map:
   item:

 anothermap:
 ```
 
 As you can see, it still removes the newline at the start of the document, but as far as I can tell that's not really a problem.
 
 
I hope that the test case is sufficient and that it's in the right place, I wasn't really sure where to put it.

### DIFF
--- a/src/stringify/stringifyPair.ts
+++ b/src/stringify/stringifyPair.ts
@@ -127,7 +127,7 @@ export function stringifyPair(
       ws += `\n${indentComment(cs, ctx.indent)}`
     }
     if (valueStr === '' && !ctx.inFlow) {
-      if (vsb && !vcb && valueComment) ws = '\n\n'
+      if (ws === '\n' && valueComment) ws = '\n\n'
     } else {
       ws += `\n${ctx.indent}`
     }

### CODE REVIEW

File: src/stringify/stringifyPair.ts [hunk 1]
```diff
@@ -127,7 +127,7 @@ export function stringifyPair(
       ws += `\n${indentComment(cs, ctx.indent)}`
     }
     if (valueStr === '' && !ctx.inFlow) {
-      if (ws === '\n') ws = '\n\n'
+      if (vsb && !vcb && valueComment) ws = '\n\n'
```
eemeli: Let's extend the previous conditional here.
```suggestion
      if (ws === '\n' && valueComment) ws = '\n\n'
```

Neutronlul: Sounds good.

Issue #643, opened 2025-11-22
Expose chomp indicator through documented API


ericsciple:
**Is your feature request related to a problem? Please describe.**

When using the `keepSourceTokens` parse option, I need to access the block scalar header information (specifically the chomp indicator: `-`, `+`, or implicit clip) to validate YAML for GitHub Actions workflows. Currently, this information is only accessible through undocumented internal structures (`srcToken.props`), which requires using `any` types and may not be stable across versions.

**Describe the solution you'd like**

I would like a documented, stable API for accessing the block scalar header, particularly the chomp indicator.

Proposed solution:
- https://github.com/eemeli/yaml/pull/644

**Describe alternatives you've considered**

Current workaround (accessing undocumented internals):
```typescript
if (token.srcToken && "source" in token.srcToken) {
  if (token.srcToken.type === "block-scalar" && "props" in token.srcToken) {
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const props = token.srcToken.props as any[];
    const headerProp = props.find((p: any) => p.type === "block-scalar-header");
    if (headerProp && "source" in headerProp) {
      blockScalarHeader = headerProp.source; // e.g., "|-", "|+", etc.
    }
  }
}
```

This works in practice but has the following issues:
1. The `props` array is not typed in TypeScript definitions
2. It's unclear if this structure is stable across versions
3. Using `any` types defeats the purpose of TypeScript
4. No indication whether this is intended to be a public API

**Additional context**

- Using `yaml` version: `^2.0.0-8` (tested with `2.8.1` - structure is the same)
- The `keepSourceTokens` option is already documented, which is great
- My use case involves validating YAML for GitHub Actions workflows where chomp indicators (`-`, `+`) affect how trailing newlines are handled in expressions
  - Refer PR: https://github.com/actions/languageservices/pull/215
- The chomping behavior itself works correctly; I just need programmatic access to know which indicator was used in the source

eemeli:
So the specific bit of info you need is whether there's an explicit `+` in the header?

ericsciple:
@eemeli yes exactly. I think that would suffice.

eemeli:
Given thow rarely this info is needed, I think the appropriate fix here isn't to add a specific new API for it, but to clarify that the [CST node structures](https://github.com/eemeli/yaml/blob/main/src/parse/cst.ts) ought to be considered a documented rather than undocumented API, and so something like this (modified from the above) is valid and supported:

```js
[...]

PR #644, opened 2025-11-22
Expose block scalar header

ericsciple:
## Summary

Adds property `blockScalarHeader` on `Scalar.Parsed` when `keepSourceTokens` is enabled. This includes the style indicator, chomp indicator, and explicit indentation. For example `|`, `|-`, `|+`, or `|2`.

Closes:
- https://github.com/eemeli/yaml/issues/643

## Usage

```typescript
const doc = YAML.parseDocument('|-\n  content\n', { keepSourceTokens: true });
console.log(doc.contents.blockScalarHeader); // '|-'
```

## Implementation

Added `blockScalarHeader?: string` to `Scalar.Parsed` interface and populated it during parsing from the CST token. Follows the same pattern as other optional metadata properties like `srcToken`.

### DIFF
--- a/src/compose/compose-scalar.ts
+++ b/src/compose/compose-scalar.ts
@@ -53,6 +53,12 @@ export function composeScalar(
   if (tagName) scalar.tag = tagName
   if (tag.format) scalar.format = tag.format
   if (comment) scalar.comment = comment
+  if (ctx.options.keepSourceTokens && token.type === 'block-scalar') {
+    const headerToken = token.props[0]
+    if (headerToken.type === 'block-scalar-header') {
+      scalar.blockScalarHeader = headerToken.source
+    }
+  }
 
   return scalar as Scalar.Parsed
 }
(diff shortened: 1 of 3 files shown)


src/compose/compose-scalar.ts
```diff
@@ -53,6 +53,12 @@ export function composeScalar(
   if (tagName) scalar.tag = tagName
   if (tag.format) scalar.format = tag.format
   if (comment) scalar.comment = comment
+  if (ctx.options.keepSourceTokens && token.type === 'block-scalar') {
+    const headerToken = token.props[0]
```
ericsciple: The access to `token.props[0]` is guaranteed to be safe by the parser structure and is consistent with existing code patterns in this repository.

When a `BlockScalar` token is created, it's [always initialized](https://github.com/eemeli/yaml/blob/92821f2b8164f9831ff5a51f6e5a575e06365742/src/parse/parser.ts#L870-L876) with the `block-scalar-header` token as the first element in `props`.

The codebase already relies on this guarantee in the same way at [`cst-scalar.ts:264`](https://github.com/eemeli/yaml/blob/92821f2b8164f9831ff5a51f6e5a575e06365742/src/parse/cst-scalar.ts#L264).

PR #651, opened 2026-01-10
feat: Expose `isMergeKey` and `MERGE_KEY` `const`ant

danielbayley:
Trying to use `isMergeKey` for something… Is there a reason it isn’t exposed alongside `isScalar` etc? Also, seems like `MERGE_KEY` should be a [shared symbol](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry), no?


src/schema/yaml-1.1/merge.ts
```diff
@@ -12,7 +12,7 @@ import type { ScalarTag } from '../types.ts'
 // Keys in mapping nodes earlier in the sequence override keys specified in
 // later mapping nodes. -- http://yaml.org/type/merge.html
 
-const MERGE_KEY = '<<'
+export const MERGE_KEY = '<<'
```
eemeli: As I mention in my previous review, I intend to make type checking stricter in v3. To that end, we shouldn't be exposing `MERGE_KEY`, but providing and enforcing the use of a new `Merge` class for this purpose.

```suggestion
const MERGE_KEY = '<<'
```

eemeli (changes requested):

PR #652, opened 2026-01-20
Fix standalone CR line break handline per YAML 1.2 spec

### LINKED ISSUE #595: Single carriage return (`\r`) not treated as line break
**Describe the bug**
Single carriage return line endings (`\r`) are treated as non-breaking characters. This is contrary to the [YAML spec](https://yaml.org/spec/1.2-old/spec.html#id2774608), which explicitly calls out CR, LF and CRLF as the (only) possible line breaks.

A similar problem afflicts other parsers, for example https://github.com/jbeder/yaml-cpp/issues/986

**To Reproduce**
```js
const lfTokens = Array.from((new Lexer()).lex("text: \"a\n\n\n\n b\""));
console.log(lfTokens.length);

const crTokens = Array.from((new Lexer()).lex("text: \"a\n\r\r\r b\""));
console.log(crTokens.length);
```

Note the output:

```
6
9
```

**Expected behaviour**
```
6
6
```

The extra tokens in the second case are a result of `\r` being treated like any other (non-breaking) character, meaning the tokens are split only at the first `\n` and not reassembled:

(..., `a`, `\n`, `\x1F`, `\r\r\r b`)

- Environment: [e.g. Node.js 14.7.0 or Chrome 87.0]
Node.js 20.17.0

- `yaml`: [e.g. 1.10.0 or 2.0.0-2]
2.6.1

**Additional context**
There are several places this seems to manifest, for example:

https://github.com/eemeli/yaml/blob/main/src/parse/lexer.ts#L223
https://github.com/eemeli/yaml/blob/main/src/parse/lexer.ts#L191

Keylan:
## Summary

Fixes #595

Treats standalone `\r` (carriage return) as a line break per the YAML 1.2 specification, which defines line breaks as `\n`, `\r\n`, or `\r`.

## Changes
- **Lexer**: Added `findLineBreak()` method and updated parsing functions to handle `\r` as a line break
- **CST**: Added `\r` case to `tokenType()`
- **Block scalar resolution**: Updated `splitLines()` regex to split on `\r`
- **Flow scalar resolution**: Updated `foldLines()` and `foldNewline()` to handle `\r`

## Notes

This PR was developed with LLM assistance.

### DIFF
--- a/src/compose/resolve-flow-scalar.ts
+++ b/src/compose/resolve-flow-scalar.ts
@@ -159,16 +159,12 @@ function doubleQuotedValue(source: string, onError: FlowScalarErrorHandler) {
       let next = source[++i]
       const cc = escapeCodes[next]
       if (cc) res += cc
-      else if (next === '\n') {
-        // skip escaped newlines, but still trim the following line
-        next = source[i + 1]
-        while (next === ' ' || next === '\t') next = source[++i + 1]
-      } else if (next === '\r' && source[i + 1] === '\n') {
-        // skip escaped CRLF newlines, but still trim the following line
+      else if (next === '\r' && source[i + 1] === '\n') {
+        // skip escaped CRLF, but still trim the following line
         next = source[++i + 1]
         while (next === ' ' || next === '\t') next = source[++i + 1]
-      } else if (next === '\r') {
-        // skip escaped standalone CR, but still trim the following line
+      } else if (next === '\n' || next === '\r') {
+        // skip escaped newline (LF or CR), but still trim the following line
         next = source[i + 1]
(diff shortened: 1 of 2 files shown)


eemeli:
Before considering this, my question from https://github.com/eemeli/yaml/issues/595#issuecomment-2544888400 remains relevant:
> While recognising that this is a divergence from the spec, does it affect any real-world use for you? As in, do you have some content that is actually using a lone `\r` as a line break?

If there's no actual content using `\r` as a line break, there's no reason to make it possible.

eemeli:
Closing, as it doesn't look like there's any real use case being served by this change.

PR #654, opened 2026-02-05
Add keepIndentStep option to parse

### LINKED ISSUE #283: Array-indent not maintained between parseDocument/toString
**Describe the bug**

First of all, thanks for an awesome yaml parser/stringifyer! I just wrote this this weekend: [yaml-diff-patch](https://github.com/grantila/yaml-diff-patch) using this library to allow modifying a yaml using JSON patches.

I'm not sure if this is considered a bug, but I'd very much hope so. The library would be most useful if it maintains as much as possible from the source yaml when parsing the document into AST/CST so that stringifying it produces the exact same result. This means that minimal (or no) modifications to the AST would produce a minimal (or no) diff when text diffing the original yaml with the modified. In other words, **only** what you change, would be part of the final diff.

**To Reproduce**

The following:

```yaml
a:
- b: c
```

when `parseDocument` -> `toString` produces:

```yaml
a:
  - b: c
```

**Expected behaviour**

I totally understand how this looks better. I also totally understand that a document with both these indentations in different places is pretty ugly, and being able to "reformat" it is great. However, being able to maintain all blocks' indentations individually as-is makes more sense when trying to make a minimal diff that only touches what it needs.

I noticed (in 1.10.2) that quite a few things get re-organized, long lines get chopped into multiple shorter, long one-line arrays get split into line-by-line arrays etc. I haven't tried that in 2.0 yet, so you can ignore that for now.

- Environment: Node.js 14.15.1
- `yaml`: 1.10.2 and 2.0.0-6

yjp20:
Adds a keepIndentStep parse option that preserves the original indentation between keys and values in block maps. When enabled, each Pair stores its original indentStep string, which stringifyPair uses instead of the context's default.

<details>
<summary>previous keepIndent option <a href="https://github.com/eemeli/yaml/commit/83326fa44b67d8af8604818072ff08afb43741fb">83326fa</a></summary>

I believe this fixes https://github.com/eemeli/yaml/issues/283 by adding a `keepIndent` option to toString.

I temporarily enabled this `keepIndent` + `keepSourceToken` on by default  and ran all the tests and it seems to work.

Alternatively, indent could be an optional property on the AST, which is maybe better in theory but also leads to some odd situations. Another alternative to this implementation might be something more general like `keepSourceFormatting` where in addition to formatting changes, we would attempt to keep more formatting e.g. in multiline flow text we wouldn't re-fold text if sourceToken exists.

This would be really helpful for us at [Stainless](https://www.stainless.com), as we apply "transforms" to a yaml file, but also need to textually diff the original yaml file against the transformed file to show changes to our users. Spurius diffs make this process really difficult to inspect.

And while I'm at it just want to say thank you for your work on this library, much of my work at Stainles would just not be possible without it. Diffing, transforming, adding comments, etc. to yaml files are a core part of what we do, and we use this library for things like just parsing YAML, analyzing source in our language server, showing great diffs in our transforms product, and so, so many more things.

</details>


yjp20:
Hey @eemeli, just getting around to this 2 months later, sorry about that.

I thought more carefully about your suggestion and I think the ability for people to change indentation however they want prorgrammatically is a pretty compelling reason to extract out the logic a property on `indentStep` and populate this in the parse step. This allows people to fine tune the behavior if they want, even if they didn't get Pair through `keepIndentStep`.

There are also weird cases where you might want to graft a Node from a parsed tree where it makes more sense to store indents as relative (rather than the absolute value encoded into the the `srcToken` which is what I was using before).

I landed on adding an `indentStep: number` property to Pair, where at the moment it's only respected on block maps. What do you think about this approach?

It is probably possible to extend this to flow nodes, but would require a bit more work in terms of preserving formatting for the fully general case there since you could have a lot more weird things happen in flow (e.g. 3 elements on each line or something) that cannot happen in block.

Also as a fair disclaimer, this was partially written by AI (basically just tests / docs).

PR #666, opened 2026-02-16
Refactor collections

eemeli:
This is a pretty deep overhaul of the document-level API's collections. Previously, these all extended an internal `Collection` class and included an `.items` array of values.

The intent with these changes is to make the collections easier to use, more performant, and to make them more like standard JS collections.

**YAMLSeq** now extends Array, and so works with `[number]` accessors. It overrides the array-mutating methods `.fill()`, `.push()`, `.splice()`, and `.unshift()` so that they can be used with values that aren't already wrapped as nodes. Setting `[number]` accessed values is not overridden, but a `.set()` method is provided for that.

**YAMLMap** is now built around an internal `.values` Map, keyed by a representation of the pair's key. This key can be user-defined with the `mapKey` option, which replaces the `uniqueKeys` comparator. It provides `.delete()`, `.get()`, `.getPair()`, `.has()`, `.keyOf()`, `.pairs()`, and `.set()` utility methods for working with its values.

**YAMLSet** no longer extends YAMLMap, but is its own class, with a very similar API as YAMLMap, with `.add()`, `.delete()`, `.get()`, `.has()`, and `.keyOf()` methods. Its `.values` does not contain Pair instance as YAMLMap does, but the set's node values directly.

The serializations of maps and sets are now distinct from each other, with maps now never leaving out the value (i.e. you'll get `!!set { foo, bar }`, but `!!map { foo: null, bar: null }`). Note that `!!set` is not enabled by default for serialization, so a Set will be represented by a sequence unless explicitly enabled:
```js
import { stringify } from 'yaml'

const set = new Set(['foo', 'bar'])
stringify(set, { flow: true }) // '[ foo, bar ]\n'
stringify(set, { customTags: ['set'], flow: true }) // '!!set { foo, bar }\n'
```

`sortMapEntries` is now a ToString option rather than a Schema option, and only affects the serialization. In addition to maps, it also affects the serialization of sets.

The path-like methods `.addIn()`, `.deleteIn()`, `.getIn()`, `.hasIn()`, and `.setIn()` are dropped from the document and collection APIs. You should instead chain `.get()` and `[]` accessors as appropriate. The document's `.add()`, `.delete()`, and `.has()` methods are dropped, adding `.getPair()`; `.get()` and `.set()` are adapted to work with the collection updates.


eemeli:
The CI failure appears to be due to something breaking in `playwright` and/or `@vitest/browser-playwirght` in the last few weeks. Rather than trying to diagnose it more deeply, I'm going to ignore it and see if they get it fixed upstream.

PR #670, opened 2026-02-27
Option to add trailing comma to multiline flow for `toString`

### LINKED ISSUE #669: Option in `ToStringOptions` to add a trailing comma to FlowSequences and FlowMaps
I work on vscode-yaml, and we're investigating using the `toString()` function from parsed YAML documents to replace the implementation of our formatting feature. Our current implementation uses prettier to format the document. For the most part, we're able to provide all the same formatting options with `toString()` as we are with prettier.

However, one of the options prettier provides is called `trailingComma`; enabling it, then formatting turns something like this:

```yaml
key:
  [
    "value1",
    "value2",
    "value3",
    "value4",
    "value5",
    "value6",
    "value7",
    "value8",
    "value9"
  ]
```

into this (note the added comma):

```yaml
key:
  [
    "value1",
    "value2",
    "value3",
    "value4",
    "value5",
    "value6",
    "value7",
    "value8",
    "value9",
  ]
```

This isn't provided as an option currently in `ToStringOptions`.

A new boolean option to `ToStringOptions` that, when enabled, adds a trailing comma when a flow sequence or flow map is split across multiple lines.

We could remove this formatting option from vscode-yaml, stick with using prettier for formatting, or write our own formatter.

**Additional context**
I'm willing to implement this functionality if it's something you're okay with having in the repo. I also understand if you don't want to add more options to the `toString()` functionality.

Here are some relevant issues where we talked about changing the implementation of vscode-yaml's formatter:
- https://github.com/redhat-developer/yaml-language-server/pull/983
- https://github.com/redhat-developer/yaml-language-server/issues/933

datho7561:
Disabled by default. When enabled, the last entry in a flow map or flow sequence will have a `,` after it if it's split across multiple lines.

eg.

```yaml
{
  a: aaa, # comment
  b: bbb
}
```

When parsed then converted back to string will become:

```yaml
{
  a: aaa, # comment
  b: bbb,
}
```

Closes #669

### DIFF
--- a/src/stringify/stringify.ts
+++ b/src/stringify/stringify.ts
@@ -55,9 +55,9 @@ export function createStringifyContext(
       nullStr: 'null',
       simpleKeys: false,
       singleQuote: null,
+      trailingComma: false,
       trueStr: 'true',
       verifyAliasOrder: true,
-      trailingComma: false,
     },
     doc.schema.toStringOptions,
     options
--- a/src/stringify/stringify.ts
+++ b/src/stringify/stringify.ts
@@ -57,7 +57,7 @@ export function createStringifyContext(
       singleQuote: null,
       trailingComma: false,
       trueStr: 'true',
-      verifyAliasOrder: true,
+      verifyAliasOrder: true
     },
     doc.schema.toStringOptions,
     options
--- a/src/stringify/stringify.ts
+++ b/src/stringify/stringify.ts
@@ -56,7 +56,8 @@ export function createStringifyContext(
       simpleKeys: false,
       singleQuote: null,
       trueStr: 'true',
-      verifyAliasOrder: true
+      verifyAliasOrder: true,
+      trailingComma: false,
     },
     doc.schema.toStringOptions,
     options
--- a/src/stringify/stringifyCollection.ts
+++ b/src/stringify/stringifyCollection.ts
@@ -108,7 +108,6 @@ function stringifyFlowCollection(
   let reqNewline = false
   let linesAtValue = 0
   const lines: string[] = []
-  let precomputedLineLength = -1
   for (let i = 0; i < items.length; ++i) {
     const item = items[i]
     let comment: string | null = null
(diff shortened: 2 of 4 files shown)


src/stringify/stringify.ts
```diff
@@ -56,7 +56,8 @@ export function createStringifyContext(
       simpleKeys: false,
       singleQuote: null,
       trueStr: 'true',
-      verifyAliasOrder: true
+      verifyAliasOrder: true,
+      trailingComma: false,
```
eemeli: Should be in alphabetic order.

tests/doc/stringify.ts
eemeli, line 1: The tests added here are pretty hard to read, and they seem pretty random with how they use the `source` template tag. Please use the `lineWidth` and `minContentWidth` options to get the behaviour you want without needing to rely on the default wrapping at 80 characters.

src/stringify/stringifyCollection.ts
```diff
@@ -134,7 +134,18 @@ function stringifyFlowCollection(
 
     if (comment) reqNewline = true
     let str = stringify(item, itemCtx, () => (comment = null))
-    if (i < items.length - 1) str += ','
+    if (i < items.length - 1) {
+      str += ','
+    } else if (ctx.options.trailingComma) {
+      // 'look forwards' to see if entries will be connected with newlines
+      const willReqNewline = reqNewline
+        || lines.reduce((sum, line) => sum + line.length + 2, 2) + (str.length + 2) + (comment ? lineComment(str, itemIndent, commentString(comment)).length : 0) > ctx.options.lineWidth
```
eemeli: This is doing a bunch of extra work that shouldn't be duplicated. For instance, there should be no need to render the comment both here and a few lines below, or for the `.reduce()` to be done both here and below. Also, this is ignoring the possibility of `lineWidth` being disabled.

eemeli (changes requested):
See inline for details, but also please rebase on top of the `v2` branch, as the CI jobs were incorrectly filtering out PRs against non-default branches, and there are at least linter errors that need fixing.

Also, plese [say so](https://github.com/eemeli/yaml/blob/main/docs/CONTRIBUTING.md#contributing-code) if you're using an LLM.

datho7561:
Thanks for the review! I'm not using AI. I'll try to get to the comments today.

tests/doc/stringify.ts
datho7561: I tried made them look more consistent and added a shorter line width for the tests that are testing that. Let me know if they look better now.

src/stringify/stringifyCollection.ts
datho7561: I added a test case for lineWidth being disabled.

I'm going to get back to this tomorrow, the logic is messy and unfortunately I ran out of time for today

datho7561:
I changed the implementation to reuse the computed values of the comment and line width calculation. The code's a bit messier than before as a result, I don't think there's a good way around this though.  I fixed the `lineWidth = 0` case.

src/stringify/stringifyCollection.ts
```diff
@@ -134,20 +135,50 @@ function stringifyFlowCollection(
...
-    if (comment) str += lineComment(str, itemIndent, commentString(comment))
-    if (!reqNewline && (lines.length > linesAtValue || str.includes('\n')))
-      reqNewline = true
-    lines.push(str)
-    linesAtValue = lines.length
+    if (i < items.length - 1) {
+      str += ','
+      if (comment) str += lineComment(str, itemIndent, commentString(comment))
+      if (!reqNewline && (lines.length > linesAtValue || str.includes('\n')))
+        reqNewline = true
```
eemeli:
```suggestion
      reqNewline ||= lines.length > linesAtValue || str.includes('\n')
```

src/stringify/stringifyCollection.ts
```diff
@@ -134,20 +135,50 @@ function stringifyFlowCollection(
...
+          newlineDueToLength = precomputedLineLength > ctx.options.lineWidth
+        }
+
+        const willReqNewline = reqNewline
+          || newlineDueToLength
+          || lines.length > linesAtValue
+          || str.includes('\n')
+        if (willReqNewline) {
+          str += ','
+        }
```
eemeli: This is carrying too much state around. If there is a `comment`, `reqNewline` will have been set true earlier, and there's no need to pre-render the comment. We also shouldn't need `newlineDueToLength`, `willReqNewline`, or a `precomputedLineLength` that's retained past here; why not set `reqNewline` directly if/when necessary?

eemeli (changes requested):

datho7561:
Didn't realize I could simplify it that way; thanks for the comments.

src/stringify/stringifyCollection.ts
```diff
@@ -134,10 +134,19 @@ function stringifyFlowCollection(
...
+      }
+      reqNewline ||= lines.length > linesAtValue || str.includes('\n')
+      if (reqNewline) {
+        str += ','
+      }
+    }
     if (comment) str += lineComment(str, itemIndent, commentString(comment))
-    if (!reqNewline && (lines.length > linesAtValue || str.includes('\n')))
-      reqNewline = true
+    reqNewline ||= lines.length > linesAtValue || str.includes('\n')
```
eemeli: Let's move this up to just after the `let str = ...` line; then it doesn't need to be duplicated within the if statement above.

eemeli (changes requested):
One small fix inline, and could you run `npm run prettier` to fix the code style? I thought that check was included in the CI's lint action, but apparently it's not.

eemeli (approved):
🎉

PR #678, opened 2026-04-28
Use the null prototype when parsing YAML maps

### LINKED ISSUE #675: Use a null prototype to represent YAML maps
YAML defines maps. YAML maps map really well to the JavaScript `Map` type, but this is often inconvenient to use in practice. So I think the current default to use a plain object by default makes sense.

The problem however, is that the default `Object` prototype has some default properties. This may lead to bugs, such as prototype pollution. While I don’t expect to see this happen inside of this package, it might happen in user code that consumes the values used with this package.

Use the `null` prototype for parsed YAML maps.

N/A

**Additional context**
- https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects
- https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/Prototype_pollution#create_javascript_objects_with_null_prototype
- This is also used in other places where a plain object represents a map, such as [RegExp capturing groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#groups).

remcohaszing:
YAML maps map a key to a value. These keys may conflict with keys on the JavaScript `Object` prototype. Several JavaScript APIs that use a plain object to represent a key/value map, use a null prototype.

Closes #675

### DIFF
--- a/src/nodes/Pair.ts
+++ b/src/nodes/Pair.ts
@@ -36,7 +36,7 @@ export class Pair<
     doc: Document<DocValue, boolean>,
     ctx: ToJSContext
   ): ReturnType<typeof addPairToJSMap> {
-    const pair = ctx.mapAsMap ? new Map() : {}
+    const pair = ctx.mapAsMap ? new Map() : { __proto__: null }
     return addPairToJSMap(doc, ctx, pair, this)
   }
 
--- a/src/nodes/Pair.ts
+++ b/src/nodes/Pair.ts
@@ -36,7 +36,7 @@ export class Pair<
     doc: Document<DocValue, boolean>,
     ctx: ToJSContext
   ): ReturnType<typeof addPairToJSMap> {
-    const pair = ctx.mapAsMap ? new Map() : { __proto__: null }
+    const pair = ctx.mapAsMap ? new Map() : Object.create(null)
     return addPairToJSMap(doc, ctx, pair, this)
   }
 
--- a/src/nodes/Pair.ts
+++ b/src/nodes/Pair.ts
@@ -36,7 +36,9 @@ export class Pair<
     doc: Document<DocValue, boolean>,
     ctx: ToJSContext
   ): ReturnType<typeof addPairToJSMap> {
-    const pair = ctx.mapAsMap ? new Map() : Object.create(null)
+    const pair = ctx.mapAsMap
+      ? new Map()
+      : (Object.create(null) as Record<string, unknown>)
     return addPairToJSMap(doc, ctx, pair, this)
   }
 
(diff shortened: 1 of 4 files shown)


remcohaszing:
I realized `JSON.parse()` doesn’t use null prototypes, but for example the [`toml`](https://github.com/BinaryMuse/toml-node) package goes.

src/nodes/Pair.ts
```diff
@@ -36,7 +36,7 @@ export class Pair<
     doc: Document<DocValue, boolean>,
     ctx: ToJSContext
   ): ReturnType<typeof addPairToJSMap> {
-    const pair = ctx.mapAsMap ? new Map() : {}
+    const pair = ctx.mapAsMap ? new Map() : { __proto__: null }
```
eemeli: Directly setting `__proto__` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto), and not the way we should do this. Please use `Object.create(null)` instead.

eemeli (changes requested):

src/nodes/Pair.ts
remcohaszing: Interesting! TIL!

eemeli (changes requested):
There's a linter failure, otherwise looks good!

src/nodes/Pair.ts
remcohaszing: Actually, I see that article states:

> The `__proto__` property can also be used in an object literal definition to set the object `[[Prototype]]` on creation, as an alternative to [`Object.create()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create). See: [object initializer / literal syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer). That syntax is standard and optimized for in implementations, and quite different from `Object.prototype.__proto__`.

And the linked [Object initializer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#prototype_setter) page even uses it in examples.

So I think my initial implementation was good, and slightly more readable.

eemeli:
The documentation will also need an update, at least under [Document#toJS()](https://eemeli.org/yaml/#document-methods).

remcohaszing:
A friend of mine did some benchmarking on null prototypes and figured that objects with a null prototype provide a 3x memory penalty and a significant access penalty vs regular objects, because of V8 optimizations.

So maybe it’s better to not merge this PR altogether.

eemeli:
Yeah, let's not do this if it introduces a performance penalty -- and as the `mapAsMap` option is already available when conflicts with Object prototype methods are a concern.

PR #679, opened 2026-04-28
Convert the bin entry to TypeScript

remcohaszing:
This means it’s type checked. There’s no need to have this particular file as JavaScript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   ],
   "type": "module",
   "main": "./dist/index.js",
-  "bin": "./bin.js",
+  "bin": "./dist/bin.js",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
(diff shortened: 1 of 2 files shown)


eemeli (changes requested):
Why? There's no benefit to this change. Note how even the updated contents are valid JS.

remcohaszing:
I don’t see a reason to treat the `bin` entry different from any other part of the source code. With this change, the file is both type checked and linted.

This just seemed like a small improvement I noticed while looking at the code base.

eemeli:
The `bin.js` file needs to be executable. I find it much cleaner to have that be always set, rather than needing to add a post-build step for it (which this PR is currently missing, btw).

remcohaszing:
The `bin.js` file doesn’t need to be executable. Package managers will make sure the `bin` entries are executable on install. But I do agree it would be nicer if it is.

Both the benefits and drawbacks of this change are very minimal. It appears you don’t want this change, so I’ll just close the PR.

PR #684, opened 2026-06-01
fix: resolve out-of-range implicit floats as strings, not Infinity

### LINKED ISSUE #657: Can't round-trip out-of-range floating point
**Describe the bug**

**To Reproduce**
```js
{ gitsha: Infinity }
```

**Expected behaviour**

- Environment: NodeJS 24.12.0
- `yaml`: 2.8.1

**Additional context**

viraatdas:
Fixes #657

## Problem
An untagged scalar like `61e9540` is matched by the implicit float regex and resolved to `Infinity`, even though it cannot round-trip as a JS number. Per YAML 1.2, an implicit tag should only be applied if the value round-trips properly.

## Fix
In `findScalarTagByTest` (`src/compose/compose-scalar.ts`), reject the implicit float tag when the resolved value is non-finite, falling through to the string tag. Explicit `!!float` is unaffected, and `.inf` / `.nan` literals still resolve correctly.

## Tests
Added regression tests (covering the core and compat schema paths) that fail before and pass after.

### DIFF
--- a/src/compose/compose-scalar.ts
+++ b/src/compose/compose-scalar.ts
@@ -1,7 +1,7 @@
 import { Scalar } from '../nodes/Scalar.ts'
 import type { BlockScalar, FlowScalar, SourceToken } from '../parse/cst.ts'
 import type { Schema } from '../schema/Schema.ts'
-import type { ScalarTag } from '../schema/types.ts'
+import type { CollectionTag, ScalarTag } from '../schema/types.ts'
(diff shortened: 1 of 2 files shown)


viraatdas:
Closing this one. On a closer read of #657, warning on overflow is the direction you preferred rather than resolving overflowing scalars to strings, so this approach is not the right fit. Apologies for the noise.

PR #685, opened 2026-06-09
fix: limit recursive merge aliases

### LINKED ISSUE #677: Circular merge aliases may cause RangeError despite maxAliasCount during JS conversion
**Describe the bug**

Converting a YAML document with a cyclic merge alias to native JavaScript may throw a `RangeError: Maximum call stack size exceeded`, even when `maxAliasCount` is set.

This appears to affect aliases reached through YAML merge handling, where the usual `maxAliasCount` handling does not appear to stop recursive merge resolution before stack overflow.

**To Reproduce**
```javascript
const { parse } = require('yaml')
parse('&A { <<: *A }', { maxAliasCount: 0, merge: true })
```
Output
```
RangeError: Maximum call stack size exceeded
    at Object.visit (/poc/node_modules/yaml/dist/visit.js:38:15)
    at Alias.resolve (/poc/node_modules/yaml/dist/nodes/Alias.js:30:19)
    at Scalar.addMergeToJSMap [as addToJSMap] (/poc/node_modules/yaml/dist/schema/yaml-1.1/merge.js:31:52)
    at Object.addPairToJSMap (/poc/node_modules/yaml/dist/nodes/addPairToJSMap.js:11:13)
    at YAMLMap.toJSON (/poc/node_modules/yaml/dist/nodes/YAMLMap.js:124:28)
    at mergeValue (/poc/node_modules/yaml/dist/schema/yaml-1.1/merge.js:45:27)
    at Scalar.addMergeToJSMap [as addToJSMap] (/poc/node_modules/yaml/dist/schema/yaml-1.1/merge.js:39:9)
    at Object.addPairToJSMap (/poc/node_modules/yaml/dist/nodes/addPairToJSMap.js:11:13)
    at YAMLMap.toJSON (/poc/node_modules/yaml/dist/nodes/YAMLMap.js:124:28)
    at mergeValue (/poc/node_modules/yaml/dist/schema/yaml-1.1/merge.js:45:27)
```

**Expected behaviour**

I would expect cyclic merge aliases to either be stopped by `maxAliasCount`, or to fail with a controlled error rather than recursing until stack overflow.

- Node.js v24.15.0
- yaml: reproduced with 2.8.3, 2.0.0, 1.10.3, and 1.5.0

**Additional context**

This seems related to the existing circular-reference behavior for merge handling. The difference here is that recursive merge resolution is still reached while `maxAliasCount` is set.

https://github.com/eemeli/yaml/blob/93c951b3478b4bb061d7b5227fd64f46d3f9df7f/tests/doc/anchors.ts#L458-L465

pupuking723:
Fixes #677.

This makes cyclic merge aliases fail through the existing alias-count guard instead of recursing until the call stack overflows.

The change keeps an existing anchor entry's count when the same node is re-entered during toJS() conversion, while allowing the resolved JS value to be updated. Merge alias handling now also runs alias toJS() before resolving the merge source, so maxAliasCount is applied to aliases reached through << merge handling.

Verification:

- npm test -- tests/doc/anchors.ts -t "circular reference"
- npm test -- tests/doc/anchors.ts
- npm test
- npm run lint
- npm run test:types
- npm run build

### DIFF
--- a/src/nodes/toJS.ts
+++ b/src/nodes/toJS.ts
@@ -17,7 +17,9 @@ export class ToJSContext {
   }
 
   setAnchor(node: Node, res: unknown): void {
-    this.anchors.set(node, { aliasCount: 0, count: 1, res })
+    const anchor = this.anchors.get(node)
+    if (anchor) anchor.res = res
+    else this.anchors.set(node, { aliasCount: 0, count: 1, res })
   }
 }
 
--- a/src/nodes/toJS.ts
+++ b/src/nodes/toJS.ts
@@ -1,3 +1,4 @@
+import type { Document } from '../doc/Document.ts'
 import type { ToJSOptions } from '../options.ts'
 import type { Node } from './types.ts'
 
@@ -20,6 +21,33 @@ export class ToJSContext {
     const anchor = this.anchors.get(node)
     if (anchor) anchor.res = res
     else this.anchors.set(node, { aliasCount: 0, count: 1, res })
+  }
+
+  resolveAlias(
+    doc: Document,
+    source: Node,
+    getAliasCount: () => number
+  ): unknown {
+    let data = this.anchors.get(source)
+    if (!data) {
+      source.toJS(doc, this)
+      data = this.anchors.get(source)
+    }
+    /* istanbul ignore if */
+    if (data?.res === undefined) {
+      const msg = 'This should not happen: Alias anchor was not resolved?'
+      throw new ReferenceError(msg)
+    }
+    if (this.maxAliasCount >= 0) {
+      data.count += 1
+      data.aliasCount ||= getAliasCount()
+      if (data.count * data.aliasCount > this.maxAliasCount) {
+        const msg =
+          'Excessive alias count indicates a resource exhaustion attack'
+        throw new ReferenceError(msg)
+      }
+    }
+    return data.res
   }
 }
 
--- a/src/schema/yaml-1.1/merge.ts
+++ b/src/schema/yaml-1.1/merge.ts
@@ -1,8 +1,8 @@
 import type { Document, DocValue } from '../../doc/Document.ts'
-import { Alias } from '../../nodes/Alias.ts'
+import { Alias, getAliasCount } from '../../nodes/Alias.ts'
 import { Scalar } from '../../nodes/Scalar.ts'
 import type { ToJSContext } from '../../nodes/toJS.ts'
-import { type MapLike, YAMLMap } from '../../nodes/YAMLMap.ts'
+import type { MapLike, YAMLMap } from '../../nodes/YAMLMap.ts'
 import type { ScalarTag } from '../types.ts'
 
 // If the value associated with a merge key is a single mapping node, each of
@@ -48,9 +48,8 @@ export function addMergeToJSMap(
   map: MapLike,
   value: unknown
 ): void {
-  const source = ctx && value instanceof Alias ? value.resolve(doc, ctx) : value
-  if (Array.isArray(source) && !(source instanceof YAMLMap)) {
-    if (value instanceof Alias) value.toJS(doc, ctx)
+  const source = getMergeSource(doc, ctx, value)
+  if (Array.isArray(source)) {
     for (const it of source) mergeValue(doc, ctx, map, it)
   } else {
     mergeValue(doc, ctx, map, value)
@@ -63,8 +62,7 @@ function mergeValue(
   map: MapLike,
   value: unknown
 ) {
-  if (value instanceof Alias) value.toJS(doc, ctx)
-  const source = value instanceof Alias ? value.resolve(doc, ctx) : value
+  const source = getMergeSource(doc, ctx, value)
   const srcMap = (source as YAMLMap).toJS(doc, ctx, Map<any, any>)
   if (!(srcMap instanceof Map))
     throw new Error('Merge sources must be maps or map aliases')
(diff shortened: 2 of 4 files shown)


src/schema/yaml-1.1/merge.ts
```diff
@@ -48,9 +48,10 @@ export function addMergeToJSMap(
   map: MapLike,
   value: unknown
 ): void {
-  value = ctx && value instanceof Alias ? value.resolve(doc, ctx) : value
-  if (Array.isArray(value) && !(value instanceof YAMLMap)) {
-    for (const it of value) mergeValue(doc, ctx, map, it)
+  const source = ctx && value instanceof Alias ? value.resolve(doc, ctx) : value
+  if (Array.isArray(source) && !(source instanceof YAMLMap)) {
```
eemeli: Let's clean this up while we're here; the second condition will never be false, as YAMLMap no longer extends Array.

```suggestion
  if (Array.isArray(source)) {
```

src/schema/yaml-1.1/merge.ts
```diff
@@ -48,9 +48,10 @@ export function addMergeToJSMap(
...
+    if (value instanceof Alias) value.toJS(doc, ctx)
```
eemeli: Why is this only needed when `source` is a YAMLSeq?

eemeli (changes requested):
Calling `toJS()` for the side effect of applying the anchor node count validation is a bit non-intuitive. Is there potentially a solution in adding a method on `ToJSContext` where we move the code that's now for that in `Alias.toJS()`, and call it also from elsewhere?

pupuking723:
Updated in 5c8bb6e.

The alias-count validation now lives in `ToJSContext.resolveAlias()`, and the merge-tag path reuses that helper instead of calling `Alias.toJS()` for the side effect.

I also added coverage for merge aliases that resolve to sequences so that branch stays exercised directly.

Verification:
- `npm test -- tests/doc/anchors.ts`
- `npm test`
- `npm run lint -- src/nodes/Alias.ts src/nodes/toJS.ts src/schema/yaml-1.1/merge.ts tests/doc/anchors.ts`
- `npm run test:types`
- `npm run build`

src/schema/yaml-1.1/merge.ts
pupuking723: Adjusted this so merge aliases use the same `ToJSContext.resolveAlias(doc, source, ...)` path as `Alias.toJS()`. The anchor resolution side effect is now encapsulated in the context helper rather than handled only for YAMLSeq merge sources.

src/schema/yaml-1.1/merge.ts
pupuking723: Fixed by simplifying the branch to `if (Array.isArray(source))` now that `YAMLMap` is no longer array-like.

pupuking723:
The alias-count validation is now centralized in `ToJSContext.resolveAlias()`, and both `Alias.toJS()` and YAML 1.1 merge handling use that path. The branch also includes focused coverage for recursive merge aliases. Ready for another look when you have time.

src/nodes/toJS.ts
```diff
@@ -17,7 +18,36 @@ export class ToJSContext {
...
-    this.anchors.set(node, { aliasCount: 0, count: 1, res })
+    const anchor = this.anchors.get(node)
+    if (anchor) anchor.res = res
+    else this.anchors.set(node, { aliasCount: 0, count: 1, res })
+  }
+
+  resolveAlias(
+    doc: Document,
+    source: Node,
+    getAliasCount: () => number
```
eemeli: There's no need for `getAliasCount` to get passed around as a callback. Its definition in `src/nodes/Alias.ts` should be moved here, as this is the only place it's actually called, and its arguments simplified to drop the `anchors` argument -- it's always `this.anchors` from here.

eemeli (changes requested):
Hey, finally got back to this. The fix is starting to look pretty good; see inline for a further simplification that ought to be included.

I merged your branch from `main` to resolve a conflict from merging another PR.

pupuking723:
Moved the alias counting logic into `ToJSContext` and removed the callback and `anchors` plumbing. Both `Alias.toJS()` and the merge path now call `resolveAlias(doc, source)` directly. Full tests, types, lint, formatting, and build pass locally.

src/nodes/toJS.ts
```diff
@@ -17,7 +20,52 @@ export class ToJSContext {
...
+      if (data.count * data.aliasCount > this.maxAliasCount) {
+        const msg =
+          'Excessive alias count indicates a resource exhaustion attack'
+        throw new ReferenceError(msg)
+      }
+    }
+    return data.res
+  }
+
+  private getAliasCount(doc: Document, node: Node | Pair | null): number {
```
eemeli: Let's make this actually private.
```suggestion
  #getAliasCount(doc: Document, node: Node | Pair | null): number {
```

src/schema/yaml-1.1/merge.ts
```diff
@@ -88,3 +88,20 @@ function mergeValue(
...
+  ctx: ToJSContext,
+  value: unknown
+) {
+  if (!(value instanceof Alias)) return value
+
+  const source = value.resolve(doc, ctx)
+  if (!source) {
+    const msg = `Unresolved alias (the anchor must be set before the alias): ${value.source}`
+    throw new ReferenceError(msg)
+  }
```
eemeli, line 103: Sorry, should've caught this earlier: this needs to be tested.

eemeli (changes requested):

src/nodes/toJS.ts
pupuking723: Changed this to a JavaScript private method in ce443be.

src/schema/yaml-1.1/merge.ts
pupuking723: Added a maxAliasCount: 0 assertion to the existing single-map merge alias test in ce443be. This now exercises the merge-path resolveAlias call directly.

eemeli (approved):
Bah, I didn't notice your last update until now. 🤦

PR #688, opened 2026-06-13
fix: don't auto-detect exponential floats that overflow to ±Infinity

### LINKED ISSUE #657: Can't round-trip out-of-range floating point

naveentehrpariya:
Closes #657

## Problem

An untagged scalar like `61e9540` or (from the issue) `gitsha: 61e9540` matches the float-EXP regex, so `parseFloat()` is called — returning `Infinity`. The YAML 1.2 spec §2.4 states:

> A YAML processor may use such a type for floating-point numbers, **as long as they round-trip properly.**

`61e9540 → Infinity → .inf` is not a round-trip, so the value must not be auto-tagged as a float.

## Solution

Added a `matchDefault` predicate to `ScalarTag` that is evaluated **only during automatic tag detection** (i.e. when no explicit tag is present). The `floatExp` tag in both the core and YAML 1.1 schemas sets `matchDefault` to reject values for which `parseFloat()` returns a non-finite result, letting them fall through to the string tag.

Explicit `!!float` tags are **unaffected**: they still resolve to `Infinity` when the exponent overflows, which matches the behaviour of other YAML libraries (e.g. PyYAML).

```js
// Before
YAML.parse('gitsha: 61e9540')   // { gitsha: Infinity }  ❌
YAML.parse('value: -61e9540')   // { value: -Infinity }  ❌

// After
YAML.parse('gitsha: 61e9540')   // { gitsha: '61e9540' }  ✅
YAML.parse('value: -61e9540')   // { value: '-61e9540' }  ✅

// Explicit tag still resolves as float
YAML.parse('value: !!float 61e9540')  // { value: Infinity }  ✅ (unchanged)

// Normal finite exponentials unchanged
YAML.parse('value: 6.1e3')  // { value: 6100 }  ✅
```

## Changes

- `src/schema/types.ts` — new optional `matchDefault` on `ScalarTag`
- `src/compose/compose-scalar.ts` — `findScalarTagByTest` now also checks `matchDefault` when present
- `src/schema/core/float.ts` — `floatExp.matchDefault` guards against overflow
- `src/schema/yaml-1.1/float.ts` — same for YAML 1.1 floatExp
- `tests/doc/types.ts` — two new tests covering core and YAML 1.1 schemas

### DIFF
--- a/src/compose/compose-scalar.ts
+++ b/src/compose/compose-scalar.ts
@@ -98,12 +98,18 @@ function findScalarTagByTest(
     (schema.tags.find(
       tag =>
         (tag.default === true || (atKey && tag.default === 'key')) &&
-        tag.test?.test(value)
+        tag.test?.test(value) &&
+        (!tag.matchDefault || tag.matchDefault(value))
     ) as ScalarTag) || schema.scalar
 
   if (schema.compat) {
     const compat =
-      schema.compat.find(tag => tag.default && tag.test?.test(value)) ??
+      schema.compat.find(
+        tag =>
+          tag.default &&
+          tag.test?.test(value) &&
+          (!tag.matchDefault || tag.matchDefault(value))
+      ) ??
       schema.scalar
     if (tag.tag !== compat.tag) {
       const ts = directives.tagString(tag.tag)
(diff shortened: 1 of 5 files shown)


eemeli:

Issue #687, opened 2026-06-10
`toString()` incorrectly converts comment to string


RedCMD:
**Describe the bug**
~~similar to https://github.com/eemeli/yaml/issues/686~~
`toString()` incorrectly converts comment to string when at the start of a block-scalar with a indent indicator
and multiple newline lines are added?

**To Reproduce**
```yaml
|5
#comment
```
`parsedDocumentYaml.toString()`

**Expected behaviour**
comment to stay a comment

**Versions:**

- Environment: VSCode 1.121.0 Windows
- `yaml`: 2.7.1

**Additional context**

<img width="211" height="351" alt="Image" src="https://github.com/user-attachments/assets/cfa9ff23-252b-43e5-bb55-88605c2657f5" />

dyk1454683243-sudo:
Taking this: empty document-root block scalar (`|` / `>`) followed by a document comment should round-trip — `toString()` needs to keep an indent indicator so the comment stays a comment, not scalar content.

There is no open Fixes PR (prior #689 / #697 are closed). I'll keep the change root-only and minimal.

dyk1454683243-sudo:
Opened https://github.com/eemeli/yaml/pull/721 for this.

PR #689, opened 2026-06-14
fix: add explicit indent indicator to empty block scalar when followed by a document comment

naveentehrpariya:
## Problem

An empty block scalar whose document has a trailing comment produces a `toString()` output that does not round-trip correctly.

**Reproducer (issue #687):**
```ts
import { parseDocument } from 'yaml'

const doc = parseDocument('|5\n#comment')
// |5 → literal block scalar, explicit indent 5
// #comment is at indent 0 (< 5) → YAML comment, not block content
console.log(doc.toJSON())       // ''  ✓

const str = doc.toString()
console.log(str)                // '|\n\n\n#comment\n'  ← bug

console.log(parseDocument(str).toJSON())  // '\n\n#comment\n'  ← should be ''
```

**What goes wrong:** `stringifyDocument` sets `ctx.forceBlockIndent = true` when a document comment exists (so that the comment can be placed after the block scalar). For non-empty block scalars, `forceBlockIndent` causes a two-space `indent` in the header, which correctly pushes the comment outside the scalar's content. But the early-return path for empty strings (`if (!value) return literal ? '|\n' : '>\n'`) ignores `indent`, emitting a bare `|` with no explicit indent indicator.

After the document comment is appended, the output looks like:

```
|       ← no indent indicator → content determined by first non-empty line
        ← empty line
        ← empty line
#comment  ← indent 0, becomes the first content line → value is "\n\n#comment\n"
```

On re-parse, `#comment` is block-scalar **content**, not a YAML comment — corrupting the round-trip.

## Fix

When `indent` is non-empty (i.e. `forceBlockIndent` or a document-marker-safe context requires it), emit `|1` / `>1` instead of `|` / `>` for an empty block scalar. The explicit indent indicator `1` means content must start at indent ≥ 1; the comment at indent 0 therefore falls outside the scalar and is correctly interpreted as a YAML comment on re-parse.

```diff
- if (!value) return literal ? '|\n' : '>\n'
+ if (!value) return literal ? (indent ? '|1\n' : '|\n') : (indent ? '>1\n' : '>\n')
```

## Tests

- Regression test added for #687
- 244/244 stringify tests pass (up from 243)
- 956/956 total tests pass

### DIFF
--- a/src/compose/compose-scalar.ts
+++ b/src/compose/compose-scalar.ts
@@ -98,12 +98,18 @@ function findScalarTagByTest(
     (schema.tags.find(
       tag =>
         (tag.default === true || (atKey && tag.default === 'key')) &&
-        tag.test?.test(value)
+        tag.test?.test(value) &&
+        (!tag.matchDefault || tag.matchDefault(value))
     ) as ScalarTag) || schema.scalar
 
   if (schema.compat) {
     const compat =
-      schema.compat.find(tag => tag.default && tag.test?.test(value)) ??
+      schema.compat.find(
+        tag =>
+          tag.default &&
+          tag.test?.test(value) &&
+          (!tag.matchDefault || tag.matchDefault(value))
+      ) ??
       schema.scalar
     if (tag.tag !== compat.tag) {
       const ts = directives.tagString(tag.tag)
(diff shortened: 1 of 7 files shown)


eemeli:

PR #690, opened 2026-06-14
fix: prevent IMPOSSIBLE error for flow collection key after empty line (#646)

### LINKED ISSUE #646: Special code cause `IMPOSSIBLE` error
**Describe the bug**

An `IMPOSSIBLE` error was thrown when parsing the following code

```yaml
? - Detroit Tigers
  - Chicago cubs
: - 2001-07-23

[New York Yankees, Atlanta Braves]: [2001-07-02, 2001-08-12, 2001-08-14]
```

**To Reproduce**
Steps to reproduce the behaviour.

Call the `parse` function

```
require('yaml').parse(fs.readFileSync('./test.yaml','utf8'))
Uncaught YAMLParseError: Map comment with trailing content at line 5, column 1:


[New York Yankees, Atlanta Braves]: [2001-07-02, 2001-08-12, 2001-08-14]
^

    at Composer.onError (<CWD>\node_modules\yaml\dist\compose\composer.js:70:34)
    at Object.resolveBlockMap (<CWD>\node_modules\yaml\dist\compose\resolve-block-map.js:112:9)
    at resolveCollection (<CWD>\node_modules\yaml\dist\compose\compose-collection.js:13:27)
    at Object.composeCollection (<CWD>\node_modules\yaml\dist\compose\compose-collection.js:59:16)
    at Object.composeNode (<CWD>\node_modules\yaml\dist\compose\compose-node.js:33:38)
    at Object.composeDoc (<CWD>\node_modules\yaml\dist\compose\compose-doc.js:35:23)
    at Composer.next (<CWD>\node_modules\yaml\dist\compose\composer.js:150:40)
    at next (<anonymous>)
    at Composer.compose (\node_modules\yaml\dist\compose\composer.js:132:25) {
  code: 'IMPOSSIBLE',
  pos: [ 52, 53 ],
  linePos: [ { line: 5, col: 1 }, { line: 5, col: 2 } ]
```

**Expected behaviour**

Not sure

- Environment: Node.js v24
- `yaml`: 2.8.2

**Additional context**

This code is from Prettier test suite, and we are upgrading the `yaml` package https://github.com/prettier/prettier/pull/18419 , the original code is

```js
? - Detroit Tigers
  - Chicago cubs
:
  - 2001-07-23

? [ New York Yankees,
    Atlanta Braves ]
: [ 2001-07-02, 2001-08-12,
    2001-08-14 ]
```

Prettier format to the code above, and it's become unparseable. [Link](https://github.com/prettier/prettier/blob/9a2d9cbed6eb4762c13262fe27e86e5cd28406a7/tests/format/yaml/spec/__snapshots__/format.test.js.snap#L3314)

~This code seems to come from an early version of YAML spec, but I'm not sure.~

Update: Found, it's "Example 2.11 Mapping between Sequences" from [YAML spec 1.2.2](https://yaml.org/spec/1.2.2/)

I'm not sure if it's valid code or not. 

But the docs say 

> This really should not happen. If you encounter this error code, please file a bug.

So I opened.

naveentehrpariya:
Fixes #646.

## Problem

When a flow collection (`[...]` or `{...}`) is used as an implicit block map key and is preceded by one or more empty lines, the CST parser creates a spurious empty item in the block-map's `items` array. This leaves behind an item containing only newline tokens, which `resolveBlockMap` later treats as a trailing comment. When that comment position is before the final map content, the `IMPOSSIBLE` check fires:

```
if (commentEnd && commentEnd < offset)
  onError(commentEnd, 'IMPOSSIBLE', 'Map comment with trailing content')
```

Minimal reproduction:

```js
YAML.parse('1: 2\n\n[3]: 4')
// YAMLParseError: Map comment with trailing content  IMPOSSIBLE
```

## Root cause

In `blockMap`'s `default:` case, when `startBlockValue` returns a `flow-collection` and `atMapIndent` is true, the parser unconditionally pushes a new empty `{ start }` item before pushing the flow-collection onto the stack:

```ts
} else if (atMapIndent) {
  map.items.push({ start })  // always pushed
}
this.stack.push(bv)
```

However, when the current item `it` is already empty (no `sep`, `value`, or `explicitKey`, and `start` is empty), the new `{ start: [] }` item is redundant. The preceding item (`{ start: [newline, newline] }`) is left behind as a spurious entry. When the flow-collection is later popped off the stack, it is assigned to the newly pushed item — leaving the newline-only item orphaned.

## Fix

Skip the push when the current item is already empty and there are no accumulated `start` tokens. The flow-collection is then assigned to the existing item when it is popped, matching the behaviour already used for plain scalar map keys (`Object.assign(it, { key: token, sep: [] })`).

```ts
} else if (atMapIndent) {
  if (it.sep || it.value || it.explicitKey || start.length > 0) {
    map.items.push({ start })
  }
}
```

This was already identified as the general area of the bug in the issue comments.

## Tests

Added a regression test covering the minimal case from the issue:

```yaml
1: 2

[3]: 4
```

All existing tests pass.

### DIFF
--- a/src/compose/compose-scalar.ts
+++ b/src/compose/compose-scalar.ts
@@ -98,12 +98,18 @@ function findScalarTagByTest(
     (schema.tags.find(
       tag =>
         (tag.default === true || (atKey && tag.default === 'key')) &&
-        tag.test?.test(value)
+        tag.test?.test(value) &&
+        (!tag.matchDefault || tag.matchDefault(value))
     ) as ScalarTag) || schema.scalar
 
   if (schema.compat) {
     const compat =
-      schema.compat.find(tag => tag.default && tag.test?.test(value)) ??
+      schema.compat.find(
+        tag =>
+          tag.default &&
+          tag.test?.test(value) &&
+          (!tag.matchDefault || tag.matchDefault(value))
+      ) ??
       schema.scalar
     if (tag.tag !== compat.tag) {
       const ts = directives.tagString(tag.tag)
--- a/src/parse/parser.ts
+++ b/src/parse/parser.ts
@@ -699,7 +699,9 @@ export class Parser {
                 return
               }
             } else if (atMapIndent) {
-              map.items.push({ start })
+              if (it.sep || it.value || it.explicitKey || start.length > 0) {
+                map.items.push({ start })
+              }
             }
             this.stack.push(bv)
             return
--- a/src/schema/core/float.ts
+++ b/src/schema/core/float.ts
@@ -22,6 +22,9 @@ export const floatExp: ScalarTag = {
   tag: 'tag:yaml.org,2002:float',
   format: 'EXP',
   test: /^[-+]?(?:\.[0-9]+|[0-9]+(?:\.[0-9]*)?)[eE][-+]?[0-9]+$/,
+  // Only auto-detect as float when the value round-trips: overflow to ±Infinity
+  // must fall through to the string tag (YAML 1.2 spec §2.4).
+  matchDefault: str => isFinite(parseFloat(str)),
   resolve: str => parseFloat(str),
   stringify(node) {
     const num = Number(node.value)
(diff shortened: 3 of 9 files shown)


eemeli:

PR #697, opened 2026-06-25
fix: keep indent indicator on empty block scalar before a comment

### LINKED ISSUE #687: `toString()` incorrectly converts comment to string

maximilize:
## Problem

`Document.toString()` corrupts an empty block scalar that is followed by a document comment — the comment round-trips back as the scalar's content:

```js
const doc = parseDocument('|5\n#comment')
doc.toJSON() // '' — correct: #comment is at indent 0 < 5, so it's a comment, not content

const out = doc.toString() // '|\n\n\n#comment\n'
parseDocument(out).toJSON() // '\n\n#comment\n'  ← should be ''
```

Fixes #687.

## Cause

`blockString()` returns a bare `|`/`>` for an empty value, dropping the indent indicator even when `ctx.forceBlockIndent` is set — which `stringifyDocument` sets precisely when a document comment follows. Without the indicator the empty scalar has no explicit extent, so the trailing comment is absorbed as its content on re-parse.

## Fix

Emit the indent indicator for the empty-value case when `forceBlockIndent` is set, mirroring the existing `indentSize` logic a few lines below.

## Tests

Added a round-trip regression test in `tests/doc/comments.ts`. The full suite passes, including the YAML and JSON test suites — no regressions.

---
*🤖 This pull request was written with AI assistance (Claude Code).*

### DIFF
--- a/src/stringify/stringifyString.ts
+++ b/src/stringify/stringifyString.ts
@@ -201,7 +201,13 @@ function blockString(
         : type === Scalar.BLOCK_LITERAL
           ? true
           : !lineLengthOverLimit(value, lineWidth, indent.length)
-  if (!value) return literal ? '|\n' : '>\n'
+  if (!value) {
+    // An empty block scalar followed by a document comment needs an explicit
+    // indent indicator; without it the comment is parsed back as the scalar's
+    // own content rather than staying a comment (#687).
+    const indicator = ctx.forceBlockIndent ? (indent ? '2' : '1') : ''
+    return `${literal ? '|' : '>'}${indicator}\n`
+  }
 
   // determine chomping from whitespace at value end
   let chomp: '' | '-' | '+'
--- a/tests/doc/comments.ts
+++ b/tests/doc/comments.ts
@@ -79,6 +79,17 @@ describe('parse comments', () => {
       expect(doc.value.value).toBe('value')
       expect(doc.value.range).toMatchObject([4, 18, 18])
     })
+
+    test('empty block scalar followed by comment (#687)', () => {
+      const doc = YAML.parseDocument<YAML.Scalar, false>('|5\n#comment')
+      expect(doc.value.value).toBe('')
+      expect(doc.comment).toBe('comment')
+      // Round-trip must keep the comment a comment; without an explicit indent
+      // indicator on the empty block scalar it gets parsed back as its content.
+      const rt = YAML.parseDocument<YAML.Scalar, false>(doc.toString())
+      expect(rt.value.value).toBe('')
+      expect(rt.comment).toBe('comment')
+    })
   })
 
   describe('seq entry comments', () => {


src/stringify/stringifyString.ts
```diff
@@ -201,7 +201,13 @@ function blockString(
         : type === Scalar.BLOCK_LITERAL
           ? true
           : !lineLengthOverLimit(value, lineWidth, indent.length)
-  if (!value) return literal ? '|\n' : '>\n'
+  if (!value) {
+    // An empty block scalar followed by a document comment needs an explicit
+    // indent indicator; without it the comment is parsed back as the scalar's
+    // own content rather than staying a comment (#687).
```
eemeli, line 207: Way too verbose.

src/stringify/stringifyString.ts
```diff
@@ -201,7 +201,13 @@ function blockString(
...
+    const indicator = ctx.forceBlockIndent ? (indent ? '2' : '1') : ''
```
eemeli, line 208: Why the dependence on `indent`? Wouldn't using `2` be always valid? And does the problem actually exist for indented values, or only top-level block scalars?

Please don't just feed my comment to your LLM and have it update the code based on this; the scope of the problem ought to be determined first.

src/stringify/stringifyString.ts
```diff
@@ -201,7 +201,13 @@ function blockString(
...
+    return `${literal ? '|' : '>'}${indicator}\n`
```
eemeli, line 209: These ternaries esp. within template strings make it really hard to figure out what is the actual output here -- which ends up as a 2-3 character strings.

tests/doc/comments.ts
```diff
@@ -79,6 +79,17 @@ describe('parse comments', () => {
...
+    test('empty block scalar followed by comment (#687)', () => {
+      const doc = YAML.parseDocument<YAML.Scalar, false>('|5\n#comment')
+      expect(doc.value.value).toBe('')
+      expect(doc.comment).toBe('comment')
+      // Round-trip must keep the comment a comment; without an explicit indent
+      // indicator on the empty block scalar it gets parsed back as its content.
+      const rt = YAML.parseDocument<YAML.Scalar, false>(doc.toString())
+      expect(rt.value.value).toBe('')
+      expect(rt.comment).toBe('comment')
+    })
```
eemeli, line 92: This doesn't seem to actually validate the serialization change this PR makes.

eemeli (changes requested):

maximilize:
Closing this one.

You asked for the scope first, so for the record: the round-trip only breaks for a block scalar at the document root. Nested empty ones (`a: |5`, `- |5`, `a:\n  b: |5`) already keep the comment a comment without any indicator, and for an empty scalar any indicator parses back the same — so a fix here needs much less machinery than this PR carries.

Thanks for the reviews.

PR #698, opened 2026-06-26
fix: keep newline-separated comments separate at document start (#600)

### LINKED ISSUE #600: commentBefore is null for yaml that has the comment at the beginning
**Describe the bug**

For yaml file
```yaml
# yaml-language-server: $schema=https://aka.ms/teams-toolkit/v1.7/yaml.schema.json
# Visit https://aka.ms/teamsfx-v5.0-guide for details on this file
# Visit https://aka.ms/teamsfx-actions for details on actions
version: v1.7
```
Use below js to parse it:
```js
const yaml = require("yaml")
const fs = require("fs");
const yamlContent = fs.readFileSync("example.yaml", "utf8");
const parsedYaml = yaml.parseDocument(yamlContent);
console.log(parsedYaml.commentBefore);
```

And the commentBefore is null

**To Reproduce**
Just as above

**Expected behaviour**
commentBefore should not be null

- Environment: 18
- `yaml`: 2.7.0.

**Additional context**
No

maximilize:
Fixes #600.

## Problem

At document start, comment groups separated by an empty line are merged into a single `commentBefore` on the value instead of staying separate:

```js
parseDocument('#foo\n\n#bar\n42').value.commentBefore // 'foo\n\nbar'
```

`foo` is detached from the value by an empty line, so it should belong to the document, with only `bar` on the value (as noted in the issue).

## Root cause

`parsePrelude` (`src/compose/composer.ts`) concatenated every prelude comment line into one string, and `decorate` routed that whole blob to a single node using one `afterEmptyLine` flag.

## Fix

`parsePrelude` now splits the prelude: comment groups followed by an empty line are detached into `commentBefore`, and only the final value-adjacent group stays in `comment`. `decorate` assigns the detached part to the document and the adjacent part to the value (or first collection key):

```js
const doc = parseDocument('#foo\n\n#bar\n42')
doc.commentBefore       // 'foo'
doc.value.commentBefore // 'bar'
```

Single-block and adjacent (no empty line) cases are unchanged, and round-tripping is preserved.

This also corrects the analogous collection case — the original report was a mapping — so the existing `seq comments without #` test was updated: a leading comment detached by an empty line now attaches to the document rather than the first item.

## Tests

Added a regression test in `tests/doc/comments.ts` (scalar, mapping and adjacent-comment cases). The full suite passes, and `tsc --noEmit` / `eslint` / `prettier --check` are clean.

---
*This pull request was written with AI assistance (Claude).*

### DIFF
--- a/src/compose/composer.ts
+++ b/src/compose/composer.ts
@@ -43,10 +43,21 @@ function parsePrelude(prelude: string[]) {
   let comment = ''
   let atComment = false
   let afterEmptyLine = false
+  let commentBefore = ''
   for (let i = 0; i < prelude.length; ++i) {
     const source = prelude[i]
     switch (source[0]) {
       case '#':
+        // A comment group separated from the next one by an empty line is
+        // detached from it. Keep such groups separate so that leading detached
+        // comments are assigned to the document and only the value-adjacent
+        // group to the value node, rather than merging them together (#600).
+        if (afterEmptyLine && comment !== '') {
+          commentBefore = commentBefore
+            ? `${commentBefore}\n\n${comment}`
+            : comment
+          comment = ''
+        }
         comment +=
           (comment === '' ? '' : afterEmptyLine ? '\n\n' : '\n') +
           (source.substring(1) || ' ')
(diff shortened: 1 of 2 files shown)


eemeli (changes requested):
This Claude fix is adding too much complexity, and I'm not really interested in having someone else pass my comments back and forth to their LLM to get this somewhere sensible.

PR #699, opened 2026-06-30
perf: various performance tweaks

43081j:
This is a draft for now as its basically a bunch of individual performance improvements.

Each improvement should generally be isolated in its own commit, so it is best to review by commit.

We then have two options:

1. discard the ones we don't want and merge this PR as a whole
2. pick the ones we want and open individual PRs


src/compose/compose-scalar.ts
```diff
@@ -71,7 +75,7 @@ function findScalarTagByName(
       else return tag
     }
   }
-  for (const tag of matchWithTest) if (tag.test?.test(value)) return tag
+  for (const tag of matchWithTest) if (testTag(tag.test, value)) return tag
```
fisker: The previous version is much easier to read.

src/compose/resolve-props.ts
```diff
@@ -48,7 +48,8 @@ export function resolveProps(
   let comma: SourceToken | null = null
   let found: SourceToken | null = null
   let start: number | null = null
-  for (const token of tokens) {
+  let token: SourceToken | undefined
+  for (token of tokens) {
```
fisker, line 52: This change looks bad to me. Especially for such a big block, it will be broken once the `for` loop changes to exit conditionally.

No offense. I didn't write a single line in this repo, just stopped by and expressed my personal opinion.

src/compose/compose-scalar.ts
43081j: this is basically because `test` is a `RegExp` in main. but some types are simple enough that using a `RegExp` is wasteful.

for example, `floatNaN` is a fixed set of possible values (e.g. `.nan`, `.NaN`, `.NAN`). so using equality instead of a `RegExp` is faster.

to make that possible, `test` has to be `RegExp | Func` basically.

alternatively, every `test` becomes a function and the existing ones become `(v) => pattern.test(v)`

src/compose/resolve-props.ts
43081j: it is to avoid having to do a lookup of the same value we already visited.

i could leave a comment explaining that at least. as i do agree, anyone changing this loop needs to know it has to do a full iteration

src/compose/compose-scalar.ts
```diff
@@ -88,22 +92,36 @@ function findScalarTagByName(
...
+  let tag = schema.scalar
+  for (const t of schemaTags) {
+    if (
+      (t.default === true || (atKey && t.default === 'key')) &&
+      testTag(t.test, value)
+    ) {
+      tag = t as ScalarTag
+      break
+    }
+  }
```
eemeli: Is there a performance benefit from using a `for..of` here, rather than a `schemaTags.find()`? I would find the latter easier to read.

Also, is the `as ScalarTag` really required? That seems surprising.

src/compose/resolve-flow-scalar.ts
```diff
@@ -102,44 +102,56 @@ function singleQuotedValue(source: string, onError: FlowScalarErrorHandler) {
...
-  }
-  let match = first.exec(source)
-  if (!match) return source
+  let newLineIndex = source.indexOf('\n')
+  if (newLineIndex === -1) return source
+
+  // This finds the index in the first line of the last non-space/non-tab char
+  let end = newLineIndex
+  if (source.charCodeAt(end - 1) === CHAR_CR) end--
+  let cc = source.charCodeAt(end - 1)
```
eemeli: Same also for the `line` loop below.

```suggestion
  let cc = source.charCodeAt(end - 1)
  if (cc === CHAR_CR) cc = source.charCodeAt(--end - 1)
```

src/compose/resolve-flow-scalar.ts
eemeli, line 1: I'm a bit surprised that this would be faster done manually than with regexps, but don't mind the switch.

src/nodes/addPairToJSMap.ts
```diff
@@ -11,7 +11,8 @@ export function addPairToJSMap(
   doc: Document<DocValue, boolean>,
   ctx: ToJSContext,
   map: MapLike,
-  { key, value }: Pair
+  { key, value }: Pair,
+  isPlainObject: boolean
 ): MapLike {
   if ('addToJSMap' in key) key.addToJSMap?.(doc, ctx, map, value)
```
eemeli: Should the flag also get passed here, so that it could be used by [mergeValue()](https://github.com/eemeli/yaml/blob/4ecbf764b11605e8e62ee3596f18b0a4f0f78c88/src/schema/yaml-1.1/merge.ts#L75-L80)?

src/nodes/YAMLMap.ts
```diff
@@ -147,8 +147,10 @@ export class YAMLMap<
   ) {
     ctx ??= new ToJSContext()
     const map = Type ? new Type() : ctx?.mapAsMap ? new Map() : {}
+    const isPlainObject = Type === undefined
```
eemeli: This will also be true if `ctx.mapAsMap` is true. That won't matter for the code because we only look at this value if it's false, but it's a bit confusing.

src/nodes/addPairToJSMap.ts
```diff
@@ -52,8 +52,7 @@ function stringifyKey(
   if (jsKey === null) return ''
   // eslint-disable-next-line @typescript-eslint/no-base-to-string
   if (typeof jsKey !== 'object') return String(jsKey)
-  const strCtx = createStringifyContext(doc, {})
-  strCtx.anchors = new Set()
```
eemeli, line 51: How can dropping this be valid? It's referenced only two lines below.

src/stringify/stringify.ts
```diff
@@ -29,34 +29,51 @@ export type StringifyContext = {
...
-      trueStr: 'true',
-      verifyAliasOrder: true
-    },
-    doc.schema.toStringOptions,
-    options
-  )
+  const resolved = getResolvedOptions(doc.schema)
+  const opt: StringifyContext['options'] & ToStringOptions = options
+    ? { ...resolved, ...options }
+    : resolved
```
eemeli, line 78: I have a slight concern that caching the contexts and then sometimes returning them directly is problematic, as values in the context do mutate during serialization.

src/parse/lexer.ts
```diff
@@ -13,12 +13,28 @@ function isEmpty(ch: string) {
...
+    Number.isNaN(ch) ||
+    ch === CHAR_SPACE ||
+    ch === CHAR_LF ||
+    ch === CHAR_CR ||
+    ch === CHAR_TAB
+  )
+}
+
+const flowIndicatorChars = new Set([44, 91, 93, 123, 125]) // , [ ] { }
 const invalidAnchorChars = new Set(' ,[]{}\n\r\t')
```
eemeli: The `flowIndicatorChars` should be named `flowIndicatorCharCodes`, or something similar.

src/parse/lexer.ts
```diff
@@ -86,14 +102,14 @@ class Lexer {
 
   private continueScalar(offset: number): number {
     if (this.indentNext > 0) {
-      let ch = this.source[offset]
+      let ch = this.source.charCodeAt(offset)
```
eemeli: A `ch` should always be a single-character string, and a `cc` should always be an integer character code. This applies to multiple places in this file.

src/parse/lexer.ts
```diff
@@ -514,15 +552,13 @@ class Lexer {
...
-    } while (ch === ' ' || (allowTabs && ch === '\t'))
+    let i = this.pos
+    let ch = this.source.charCodeAt(i)
+    while (ch === CHAR_SPACE || (allowTabs && ch === CHAR_TAB))
+      ch = this.source.charCodeAt(++i)
     const n = i - this.pos
     if (n > 0) {
-      const s = this.source.substr(this.pos, n)
-      this.tokens.push(s)
+      this.tokens.push(this.source.slice(this.pos, i))
```
eemeli: There's no performance benefit here, is there? So this is a stylistic change only (which is fine).

I'm aware that `.substr()` is Annex B, but I'm trying to get WinterTC to have that changed in ECMA-262, as it's implemented pretty much everywhere despite that.

src/parse/parser.ts
```diff
@@ -333,8 +338,10 @@ export class Parser {
           } else if (it.sep) {
             it.value = token
           } else {
-            Object.assign(it, { key: token, sep: [] })
-            this.onKeyLine = !it.explicitKey
+            const item = it as unknown as FilledBlockMapItem
+            item.key = token
+            item.sep = []
+            this.onKeyLine = !item.explicitKey
```
eemeli, line 344: @mgaudet Is this pattern something we're already optimising in Spidermonkey? In other words, do we actually create an intermediate object that we immediately throw away when `Object.assign()` is called with a literal object as the second argument?

src/schema/core/float.ts
```diff
@@ -2,11 +2,45 @@ import { Scalar } from '../../nodes/Scalar.ts'
...
+    case '+.Inf':
+    case '+.INF':
+    case '-.inf':
+    case '-.Inf':
+    case '-.INF':
+      return true
+    default:
+      return false
+  }
+}
```
eemeli: What's the heuristic you're using to decide whether to use an `if (value === ... || ...)` (as for nulls & bools), or this `switch` statement? And are they both better than a Set and its `.has()`?

src/schema/core/float.ts
```diff
@@ -21,7 +55,7 @@ export const floatExp: ScalarTag = {
   default: true,
   tag: 'tag:yaml.org,2002:float',
   format: 'EXP',
-  test: /^[-+]?(?:\.[0-9]+|[0-9]+(?:\.[0-9]*)?)[eE][-+]?[0-9]+$/,
+  test: (value: string) => couldBeFloat(value) && floatExpRe.test(value),
```
eemeli, line 53: Does this _really_ provide a measurable perf improvement? I would presume that non-matching values would return pretty early from the regexp test.

src/schema/types.ts
```diff
@@ -77,11 +77,11 @@ export interface ScalarTag extends TagBase {
...
    * Together with `default` allows for values to be stringified without an
-   * explicit tag and detected using a regular expression. For most cases, it's
-   * unlikely that you'll actually want to use these, even if you first think
-   * you do.
+   * explicit tag and detected using a regular expression or a test function.
+   * For most cases, it's unlikely that you'll actually want to use these,
+   * even if you first think you do.
    */
-  test?: RegExp
+  test?: RegExp | ((value: string) => boolean)
```
eemeli: Are we left with any cases that actually use a regexp, or could support for that be dropped out?

src/parse/lexer.ts
```diff
@@ -338,8 +388,7 @@ class Lexer {
     return hasIndicators
   }
 
-  private quotedScalar(): void {
-    const quote = this.charAt(0)
+  private quotedScalar(quote: string): void {
```
eemeli:
```suggestion
  private quotedScalar(quote: "'" | '"'): void {
```

eemeli (commented):
Looks pretty promising! Sorry for taking so long to get through this.

There are a few places where I'm surprised that a manual implementation is actually faster than a regexp; it'd be great to get a more accurate idea of where the performance boundaries are in modern engines.

Merging this will almost certainly be a bit messy, but maybe not too badly so?

src/compose/compose-scalar.ts
43081j: the `as` isn't needed, i think its leftover from the old code.

it changed to a `for..of` mostly to avoid allocating a function, though V8 will probably optimise away some of that anyway. happy to put it back to a `find`.

FYI the main change in this block is `testTag(t.test, value)` in that not every `test` is a `RegExp` anymore.

src/compose/resolve-flow-scalar.ts
43081j: this was one of the bigger wins. iirc mostly because we're doing numeric comparisons and jumping ahead to the `\n` rather than matching the entire string

src/nodes/addPairToJSMap.ts
43081j: a new context already has the anchors initialised as an empty set, this was just overwriting the new set with a newer set

src/nodes/addPairToJSMap.ts
43081j: i've added it, and it did seem to bump the perf a little but either way i think it makes sense for consistency

src/nodes/YAMLMap.ts
43081j: i agree it is a bit of trickery and confusion. its relying on the fact we know `addPairToJSMap` won't care about `isPlainObject` if it is a `Map`

i'll change L238 into a proper `if` and set both

src/compose/compose-scalar.ts
eemeli: I'd prefer a `.find()` here, unless that imposes a perf hit.

See https://github.com/eemeli/yaml/pull/699#discussion_r3550453879, maybe we can avoid needing to introduce `testTag`?

src/parse/lexer.ts
43081j: for whatever reason I remembered `slice` being faster than `substr` but afaik v8 takes the same path for both. so i reverted this

src/schema/core/float.ts
43081j: i basically have a benchmark repo locally where i roughly benchmark these things as i go.

my findings were:

- at some point, with enough cases, switch becomes slower (but like 50+ cases)
- switch for `string` -> `boolean` is faster than `Set<boolean>#has`
- `if` is about the same speed as the switch

but in this repo itself, it doesn't make a huge difference. my benchmark results are faster with the switch, but barely. so i'd be happy to use a `Set` instead

src/schema/core/float.ts
43081j: in my benchmarks it was ~1.5x faster for non-floats. for floats, it was about the same or slightly slower sometimes (ns worth slower).

it basically costs more to enter the regex engine than it does to do some equality checks.

in yaml benchmarks though it is about the same in most cases. so i'd be happy to discard this one

43081j: ok i re-ran my benchmarks and it is faster by ~100 ops/sec 🤔 

a few re-runs, 4100 ops/sec without `couldBeFloat` (in both locations), 4200 ops/sec with it

src/schema/types.ts
43081j: quite a few, e.g. the yaml-1.1 stuff.

but i just ran some benchmarks and it seems it'd be equal or faster if they were all functions... i.e. `(str) => re.test(str)`.

since we then don't need to interrogate the shape and can blindly `test(str)` everywhere.

i think we should probably just drop `RegExp` here

src/schema/core/float.ts
eemeli: Let's keep it, then.

src/schema/types.ts
eemeli: Let's drop the RegExp type and make this always a function.

43081j: done 👍

src/schema/json/schema.ts
```diff
@@ -57,7 +58,7 @@ const jsonScalars: ScalarTag[] = [
 const jsonError: ScalarTag = {
   default: true,
   tag: '',
-  test: /^/,
+  test: () => true,
```
eemeli, line 61: This one's definitely clearer as a function.

src/schema/common/null.ts
```diff
@@ -1,12 +1,8 @@
...
 import type { ScalarTag } from '../types.ts'
 
-const isNull = (value: string) =>
-  value === '' ||
-  value === '~' ||
-  value === 'null' ||
-  value === 'Null' ||
-  value === 'NULL'
+const nullValues = new Set(['', '~', 'null', 'Null', 'NULL'])
+const isNull = (value: string) => nullValues.has(value)
```
eemeli: Unless there's a performance reason to declare the function separately like this, I'd prefer this (and also `isBool` and `isFloatNaN`) to be inline in the ScalarTag definition.

src/schema/tags.ts
```diff
@@ -37,13 +37,13 @@ const tagsByName: {
   intOct: ScalarTag
   intTime: ScalarTag
   map: CollectionTag
-  merge: ScalarTag & { identify(value: unknown): boolean; test: RegExp }
+  merge: ScalarTag & { identify(value: unknown): boolean }
```
eemeli: The type for `merge` should include the `test` value always being set, now as a function rather than a regexp, as the base ScalarTag type only has it optionally set.

Same applies to `timestamp` here, as well as in `coreKnownTags` below.

src/compose/resolve-flow-scalar.ts
eemeli: Ah, the numeric comparisons is likely the key difference here.

src/stringify/stringify.ts
eemeli: Given that it's only the default options object that's potentially leaking through, and it's already typed as readonly, maybe this is fine? Would `Object.freeze()`ing it add a significant cost?

eemeli:
Given how I'm clearly already reviewing this, I'm going to mark this as ready for review.

eemeli (approved):
Sorry for leaving this effectively in limbo; have focused on other work recently. Let's go ahead with this, and include this in a new v3 prerelease to get a bit more real-world experience with it.

PR #700, opened 2026-07-01
fix: Parse a flow collection map key preceded by a blank line or comment

### LINKED ISSUE #646: Special code cause `IMPOSSIBLE` error

apoorva-01:
Fixes #646.

A flow collection used as an implicit block map key threw `IMPOSSIBLE` when a blank line or comment preceded it, e.g. `YAML.parse('1: 2\n\n[3]: 4')`.

Per your diagnosis: a value followed by a newline opens a new empty CST item, and when the next key is a flow collection at map indent, `blockMap`'s `atMapIndent` branch pushed a second empty item, orphaning the first mid-collection, which `resolveBlockMap` then rejects. The fix only opens a new item when the current one already has content; otherwise the flow collection assigns to the existing empty item on pop. Two regression tests (blank-line and comment) fail on `main`, pass here; suite, eslint, prettier, tsc green.

That branch is shared by block scalars and explicit keys, but only a flow collection reaches it as a new implicit key after a blank line (the others already parse fine), so the guard is effectively scoped, happy to narrow it.

Disclosure: I used an LLM to assist with investigation only. All code changes are my own. I've reviewed and understand every change, and I'm confident it's correct.

### DIFF
--- a/src/parse/parser.ts
+++ b/src/parse/parser.ts
@@ -698,7 +698,14 @@ export class Parser {
                 })
                 return
               }
-            } else if (atMapIndent) {
+            } else if (
+              atMapIndent &&
+              (it.sep || it.explicitKey || start.length > 0)
+            ) {
+              // Only open a new item when the current one already has content;
+              // otherwise reuse a preceding empty item that holds just a blank
+              // line or comment, rather than orphaning it before this key. An
+              // empty item is only valid at the very end of a block map.
               map.items.push({ start })
             }
             this.stack.push(bv)
--- a/src/parse/parser.ts
+++ b/src/parse/parser.ts
@@ -702,10 +702,6 @@ export class Parser {
               atMapIndent &&
               (it.sep || it.explicitKey || start.length > 0)
             ) {
-              // Only open a new item when the current one already has content;
-              // otherwise reuse a preceding empty item that holds just a blank
-              // line or comment, rather than orphaning it before this key. An
-              // empty item is only valid at the very end of a block map.
               map.items.push({ start })
             }
             this.stack.push(bv)
--- a/tests/doc/parse.ts
+++ b/tests/doc/parse.ts
@@ -219,6 +219,39 @@ describe('flow collection keys', () => {
     )
   })
 
+  test('flow collection key after a blank line (#646)', () => {
+    const doc = YAML.parseDocument('1: 2\n\n[3]: 4')
+    expect(doc.errors).toHaveLength(0)
+    expect(doc.value).toMatchObject(
+      _map([
+        [1, 2],
+        [_seq(3), 4]
+      ])
+    )
+  })
+
+  test('flow collection key after a comment line (#646)', () => {
+    const doc = YAML.parseDocument('1: 2\n# c\n[3]: 4')
+    expect(doc.errors).toHaveLength(0)
+    expect(doc.value).toMatchObject(
+      _map([
+        [1, 2],
+        [_seq(3), 4]
+      ])
+    )
+  })
+
+  test('flow collection key after an empty explicit key (#646)', () => {
+    const doc = YAML.parseDocument('?\n[3]: 4')
+    expect(doc.errors).toHaveLength(0)
+    expect(doc.value).toMatchObject(
+      _map([
+        [null, { key: { value: null }, value: null }],
+        [_seq(3), 4]
+      ])
+    )
+  })
+
   test('empty scalar as last flow collection value (#550)', () => {
     const doc = YAML.parseDocument<YAML.YAMLMap, false>('{c:}')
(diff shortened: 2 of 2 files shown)


src/parse/parser.ts
```diff
@@ -699,7 +699,11 @@ export class Parser {
                 return
               }
             } else if (atMapIndent) {
-              map.items.push({ start })
+              // Reuse a preceding empty item that only holds a blank line or
+              // comment, rather than leaving it orphaned before this key. An
+              // empty item is only valid at the very end of a block map.
+              if (it.sep || it.explicitKey || start.length > 0)
+                map.items.push({ start })
```
eemeli: This is probably right? The conditions should be included directly in the `else if` clause though, and I would appreciate a test for the explicit-key part of this.

eemeli (changes requested):

src/parse/parser.ts
apoorva-01: Done both. Folded the conditions into the `else if` and added a test for the explicit-key case.

On "probably right": I checked, and the `explicitKey` term is doing real work. Drop it and `?\n[3]: 4` (empty explicit key, then a flow-collection key) stops parsing as two entries and collapses into one garbled key. The new test locks that in.

src/parse/parser.ts
```diff
@@ -698,7 +698,14 @@ export class Parser {
...
               }
-            } else if (atMapIndent) {
+            } else if (
+              atMapIndent &&
+              (it.sep || it.explicitKey || start.length > 0)
+            ) {
+              // Only open a new item when the current one already has content;
+              // otherwise reuse a preceding empty item that holds just a blank
+              // line or comment, rather than orphaning it before this key. An
+              // empty item is only valid at the very end of a block map.
```
eemeli: This comment is excessive, and should be dropped. Other than that, this PR looks good!

When making this change, please don't rebase the branch, just add another commit to it.

eemeli (changes requested):

src/parse/parser.ts
apoorva-01: Done

eemeli (approved):

PR #701, opened 2026-07-01
fix: keep the source of numbers that overflow to Infinity

### LINKED ISSUE #660: Use `Scalar.source` when serialising numerical values
By default, numerical values (`!!int` and `!!float`) are represented using the native JS Number, which has limits to the precision and absolute value of what it can represent before overflowing:
```js
import { parseDocument } from 'yaml'

const doc = parseDocument('gitsha: 61e9540')
doc.get('gitsha', true)
```
```js
Scalar {
  value: Infinity,
  range: [ 8, 15, 15 ],
  source: '61e9540',
  type: 'PLAIN',
  format: 'EXP'
}
```

Note, however, that we already retain the string source of the value. When serialising the above, we should make use of it, so that we don't end up with:
```yaml
gitsha: .inf
```

maximilize:
Fixes #660.

A numeric value whose magnitude overflows the JS Number range (e.g. a git sha like `61e9540`) resolves to `Infinity` on parse but keeps its original `source`. `stringifyNumber` ignored that source and emitted `.inf`, so `gitsha: 61e9540` round-tripped to `gitsha: .inf` — silent data loss.

This emits the retained `source` in the non-finite branch, but only when it still resolves to the current value (`Number(source) === num`), so a later programmatic edit isn't masked by a stale source. Finite values are untouched — `JSON.stringify` already round-trips them exactly, so the source is never needed there.

Tests added under `number types`: the reported git-sha case, negative overflow, that literal `.inf`/`.nan` are unaffected, that a programmatic `Infinity` with no source still yields `.inf`, and that a stale source is dropped once it no longer resolves to the value. Full suite green including the YAML test matrix (2093) and JSON suite (340); `test:types` and `lint` clean.

This PR was written with AI assistance (Claude).

### DIFF
--- a/src/stringify/stringifyNumber.ts
+++ b/src/stringify/stringifyNumber.ts
@@ -3,12 +3,20 @@ import type { Scalar } from '../nodes/Scalar.ts'
 export function stringifyNumber({
   format,
   minFractionDigits,
+  source,
   tag,
   value
 }: Scalar): string {
   if (typeof value === 'bigint') return String(value)
   const num = typeof value === 'number' ? value : Number(value)
-  if (!isFinite(num)) return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
+  if (!isFinite(num)) {
+    // A number whose magnitude overflows the JS Number range (e.g. a git sha like
+    // `61e9540`) resolves to Infinity but keeps its original source. Emit that
+    // source rather than losing the value to `.inf`, as long as it still round-trips
+    // to the current value (so a later programmatic edit is not masked by stale source).
+    if (source && Number(source) === num) return source
+    return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
+  }
   let n = Object.is(value, -0) ? '-0' : JSON.stringify(value)
   if (
     !format &&
--- a/src/stringify/stringifyNumber.ts
+++ b/src/stringify/stringifyNumber.ts
@@ -1,20 +1,20 @@
 import type { Scalar } from '../nodes/Scalar.ts'
 
-export function stringifyNumber({
-  format,
-  minFractionDigits,
-  source,
-  tag,
-  value
-}: Scalar): string {
+export function stringifyNumber(
+  { format, minFractionDigits, source, tag, value }: Scalar,
+  version: '1.1' | '1.2' = '1.2'
+): string {
   if (typeof value === 'bigint') return String(value)
   const num = typeof value === 'number' ? value : Number(value)
   if (!isFinite(num)) {
-    // A number whose magnitude overflows the JS Number range (e.g. a git sha like
-    // `61e9540`) resolves to Infinity but keeps its original source. Emit that
-    // source rather than losing the value to `.inf`, as long as it still round-trips
-    // to the current value (so a later programmatic edit is not masked by stale source).
-    if (source && Number(source) === num) return source
+    // Preserve the original source of a number that overflows to Infinity (e.g.
+    // `61e9540`) as long as it still parses back to the current value.
+    if (
+      typeof value === 'number' &&
+      source &&
+      parseFloat(version === '1.1' ? source.replace(/_/g, '') : source) === num
+    )
+      return source
     return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
   }
   let n = Object.is(value, -0) ? '-0' : JSON.stringify(value)
--- a/src/stringify/stringifyNumber.ts
+++ b/src/stringify/stringifyNumber.ts
@@ -9,12 +9,10 @@ export function stringifyNumber(
   if (!isFinite(num)) {
     // Preserve the original source of a number that overflows to Infinity (e.g.
     // `61e9540`) as long as it still parses back to the current value.
-    if (
-      typeof value === 'number' &&
-      source &&
-      parseFloat(version === '1.1' ? source.replace(/_/g, '') : source) === num
-    )
-      return source
+    if (typeof value === 'number' && source) {
+      const source_ = version === '1.1' ? source.replace(/_/g, '') : source
+      if (parseFloat(source_) === num) return source
+    }
     return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
   }
   let n = Object.is(value, -0) ? '-0' : JSON.stringify(value)
--- a/src/stringify/stringifyNumber.ts
+++ b/src/stringify/stringifyNumber.ts
@@ -7,8 +7,6 @@ export function stringifyNumber(
   if (typeof value === 'bigint') return String(value)
   const num = typeof value === 'number' ? value : Number(value)
   if (!isFinite(num)) {
-    // Preserve the original source of a number that overflows to Infinity (e.g.
-    // `61e9540`) as long as it still parses back to the current value.
     if (typeof value === 'number' && source) {
       const source_ = version === '1.1' ? source.replace(/_/g, '') : source
       if (parseFloat(source_) === num) return source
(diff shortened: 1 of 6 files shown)


src/stringify/stringifyNumber.ts
```diff
@@ -3,12 +3,20 @@ import type { Scalar } from '../nodes/Scalar.ts'
...
 }: Scalar): string {
   if (typeof value === 'bigint') return String(value)
   const num = typeof value === 'number' ? value : Number(value)
-  if (!isFinite(num)) return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
+  if (!isFinite(num)) {
+    // A number whose magnitude overflows the JS Number range (e.g. a git sha like
+    // `61e9540`) resolves to Infinity but keeps its original source. Emit that
+    // source rather than losing the value to `.inf`, as long as it still round-trips
+    // to the current value (so a later programmatic edit is not masked by stale source).
+    if (source && Number(source) === num) return source
```
eemeli: This doesn't work for YAML 1.1 floats, which may use `_` internally: https://github.com/eemeli/yaml/blob/4ecbf764b11605e8e62ee3596f18b0a4f0f78c88/src/schema/yaml-1.1/float.ts#L38

Just dropping `_` characters before parsing as number isn't always valid here, because they're _not_ valid in YAML 1.2.

So we probably need another argument on this function to signal the current YAML version; if you make it optional the default (1.2) callers won't need to be changed.

Also, note that the parsers are using `parseFloat()` rather than `Number()`, which should be matched here. Furthermore, we should guard against overriding at least some (bad) user modifications of the node value by gating this behaviour on `typeof value === 'number'` being true.

Finally, the code comment here is just excessive.

tests/doc/types.ts
eemeli, line 1: Should also test that setting the value of a node that was parsed with an non-finite value is reflected in its serialization.

eemeli (changes requested):

maximilize:
All addressed in c6b2d01:

- **`version` arg** (optional, default `'1.2'`): the `yaml-1.1` number tags pass `'1.1'` so `_` separators are stripped before the round-trip check, matching the resolvers. The `stringify: stringifyNumber` references became `node => stringifyNumber(node, …)` — a tag's `stringify` is called with `(node, ctx, …)`, so a bare reference would bind `ctx` to `version`.
- `parseFloat` instead of `Number`; source is reused only when `typeof value === 'number'`, so a later value edit isn't masked by stale source.
- Trimmed the comment; added a test that editing a node parsed from a non-finite value is reflected in serialization, plus a YAML 1.1 `_`-overflow round-trip test.

Full suite green (3383, incl. the YAML/JSON matrices), tsc + prettier clean.

src/stringify/stringifyNumber.ts
```diff
@@ -1,14 +1,22 @@
...
-  if (!isFinite(num)) return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
+  if (!isFinite(num)) {
+    // Preserve the original source of a number that overflows to Infinity (e.g.
+    // `61e9540`) as long as it still parses back to the current value.
+    if (
+      typeof value === 'number' &&
+      source &&
+      parseFloat(version === '1.1' ? source.replace(/_/g, '') : source) === num
+    )
+      return source
```
eemeli:
```suggestion
    if (typeof value === 'number' && source) {
      const source_ = version === '1.1' ? source.replace(/_/g, '') : source
      if (parseFloat(source_) === num) return source
    }
```

tests/doc/types.ts
```diff
@@ -262,6 +262,48 @@ describe('number types', () => {
...
+
+    test('a programmatic Infinity with no source is emitted as .inf', () => {
+      const doc = parseDocument('n: 1')
+      doc.set('n', Infinity)
+      expect(String(doc)).toBe('n: .inf\n')
+    })
+
+    test('editing the value of a node parsed from a non-finite source is reflected', () => {
+      const doc = parseDocument('n: 61e9540') // value Infinity, source '61e9540'
+      const node = doc.get('n', true) as Scalar
```
eemeli: The v3 API doesn't have a second argument here.
```suggestion
      const node = doc.get('n') as Scalar
```

eemeli (changes requested):
Let's make it a bit easier to read.

maximilize:
Both applied in c3951aa: the guard now pulls the version-normalised value into a local `source_`, and the test uses the single-arg `doc.get('n')`. Full suite still green (3383 passing, incl. the 1.1/1.2 matrices).

tests/doc/types.ts
```diff
@@ -262,6 +262,48 @@ describe('number types', () => {
...
+      expect(String(parseDocument('n: -61e9540'))).toBe('n: -61e9540\n')
+    })
+
+    test('literal .inf / .nan are still emitted as such', () => {
+      expect(String(parseDocument('n: .inf'))).toBe('n: .inf\n')
+      expect(String(parseDocument('n: -.inf'))).toBe('n: -.inf\n')
+      expect(String(parseDocument('n: .nan'))).toBe('n: .nan\n')
+    })
+
+    test('a programmatic Infinity with no source is emitted as .inf', () => {
```
eemeli: The source isn't lost when setting the value of a scalar.
```suggestion
    test('a programmatic Infinity with finite source is emitted as .inf', () => {
```

tests/doc/types.ts
```diff
@@ -262,6 +262,48 @@ describe('number types', () => {
...
+      const node = doc.get('n') as Scalar
+      node.value = -Infinity // stale source no longer resolves to the value
+      expect(String(doc)).toBe('n: -.inf\n')
+    })
+
+    test('the source is used only while it still resolves to the value', () => {
+      const node = new Scalar(Infinity)
+      node.source = '61e9540' // overflows to Infinity → matches, so it is used
+      expect(stringifyNumber(node)).toBe('61e9540')
+      node.source = '5' // stale after a value edit: resolves to 5, not Infinity
```
eemeli: Drop the comments here.

src/stringify/stringifyNumber.ts
```diff
@@ -1,14 +1,20 @@
...
+export function stringifyNumber(
+  { format, minFractionDigits, source, tag, value }: Scalar,
+  version: '1.1' | '1.2' = '1.2'
+): string {
   if (typeof value === 'bigint') return String(value)
   const num = typeof value === 'number' ? value : Number(value)
-  if (!isFinite(num)) return isNaN(num) ? '.nan' : num < 0 ? '-.inf' : '.inf'
+  if (!isFinite(num)) {
+    // Preserve the original source of a number that overflows to Infinity (e.g.
+    // `61e9540`) as long as it still parses back to the current value.
```
eemeli: Drop the comment; it's misleading, as this case also covers NaN values.

eemeli (changes requested):
Only some prose fixes still needed, because your LLM is too verbose.

maximilize:
All three applied in 287938b: dropped the misleading comment (it also covers NaN), renamed the test to "finite source" since `set()` keeps the parsed source, and removed the inline comments. Suite still green.

tests/doc/types.ts
```diff
@@ -262,6 +262,48 @@ describe('number types', () => {
...
+    test('a programmatic Infinity with finite source is emitted as .inf', () => {
+      const doc = parseDocument('n: 1')
+      doc.set('n', Infinity)
+      expect(String(doc)).toBe('n: .inf\n')
+    })
+
+    test('editing the value of a node parsed from a non-finite source is reflected', () => {
+      const doc = parseDocument('n: 61e9540') // value Infinity, source '61e9540'
+      const node = doc.get('n') as Scalar
+      node.value = -Infinity // stale source no longer resolves to the value
```
eemeli:
```suggestion
      const doc = parseDocument('n: 61e9540')
      const node = doc.get('n') as Scalar
      node.value = -Infinity
```

eemeli (approved):

PR #704, opened 2026-07-10
fix: report C0 control characters in scalar values as errors (#703)

### LINKED ISSUE #703: No error for C0-control characters
**Describe the bug**
no error is given for C0-control characters within a string
only printable characters are allowed
https://yaml.org/spec/1.2.2/#rule-c-printable
> To ensure readability, YAML [streams](https://yaml.org/spec/1.2.2/#streams) use only the printable subset of the Unicode character set. The allowed character range explicitly excludes the C0 control block[15](https://yaml.org/spec/1.2.2/#fn:c0-block) x00-x1F (except for TAB x09, LF x0A and CR x0D which are allowed), DEL x7F, the C1 control block x80-x9F (except for NEL x85 which is allowed), the surrogate block[16](https://yaml.org/spec/1.2.2/#fn:surrogates) xD800-xDFFF, xFFFE and xFFFF.

**To Reproduce**
create a string containing any number of these characters
(I can't paste them here in GitHub)
```ts
\x00\x01\x02\x03\x04\x05\x06\x07\x08\x0b\x0c\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\e\x1c\x1d\x1e\x1f
```

**Expected behaviour**
A warning to be provided saying the non-printable c0-control characters are not allowed

**Versions:**

- Environment: VSCode 1.121.0
- `yaml`: 2.9.0

**Additional context**

Image

Image

wvwoo:
## Summary

Literal C0 control characters (other than the permitted `\t`, `\n` and `\r`) are
excluded from the YAML 1.2 `c-printable` production, but the composer accepted
them silently in plain, quoted and block scalars.

Fixes #703.

## Changes

- `src/compose/util-check-printable.ts`: new `checkPrintableChars` helper — scans
  the raw scalar **source** and emits a `CONTROL_CHAR` error for each disallowed C0
  control. It runs on the raw source rather than the resolved value, so escape
  sequences such as `\t` or `\x07` in double-quoted scalars remain valid.
- `src/compose/resolve-flow-scalar.ts` / `src/compose/resolve-block-scalar.ts`:
  wire in the check for plain/single/double-quoted and block scalars.
- `src/errors.ts`: add the `CONTROL_CHAR` error code.
- `tests/doc/errors.ts`: new tests covering control chars in each scalar style,
  plus regression guards for escaped controls and permitted whitespace (tab/NEL).

## Scope note

Scoped to C0 controls only (`\x00–\x08`, `\x0b`, `\x0c`, `\x0e–\x1f`), matching the
issue title. DEL (`\x7f`) and C1 controls are intentionally left accepted so the
official JSON test suite (`y_string_with_del_character`,
`y_string_unescaped_char_delete`) keeps passing — JSON permits unescaped DEL and
this parser is deliberately JSON-compatible. Happy to extend further if you'd
rather also revisit those JSON-compat cases.

## Testing

Full suite passes: `npx vitest run` — 3386 tests passed, 11 skipped, 0 failed,
including the official `yaml-test-suite` and `JSONTestSuite` submodules.
`tsc --noEmit`, `eslint`, and `prettier --check` are all clean. The 5 new
"should error" tests were confirmed to fail on the pre-fix code and pass with
the fix; the 2 regression-guard tests pass throughout.

## LLM use disclosure

Per this project's CONTRIBUTING.md, disclosing as required: this PR was produced
by an autonomous coding agent (Claude). A human reviewed the diff and test results
before submission.

### DIFF
--- a/src/compose/util-check-printable.ts
+++ b/src/compose/util-check-printable.ts
@@ -1,0 +1,39 @@
+import type { ComposeErrorHandler } from './composer.ts'
+
+/**
+ * C0 control characters that are excluded from the YAML 1.2 `c-printable`
+ * production, i.e. every C0 control other than the allowed `\t` (#x9),
+ * `\n` (#xA) and `\r` (#xD).
+ *
+ * See https://yaml.org/spec/1.2.2/#rule-c-printable
+ */
+const controlCharRe = /[\x00-\x08\x0b\x0c\x0e-\x1f]/g
+
+/**
+ * Reports a `CONTROL_CHAR` error for each C0 control character found in
+ * `source`, positioned at `offset + index`.
+ *
+ * The check is applied to the raw scalar source rather than its resolved
+ * value, so escape sequences in double-quoted scalars (e.g. `\t` or `\x07`)
+ * are not affected.
+ */
+export function checkPrintableChars(
+  source: string,
+  offset: number,
+  onError: ComposeErrorHandler
+): void {
+  controlCharRe.lastIndex = 0
+  let match: RegExpExecArray | null
+  while ((match = controlCharRe.exec(source)) !== null) {
+    const hex = match[0]
+      .charCodeAt(0)
+      .toString(16)
+      .padStart(2, '0')
+      .toUpperCase()
+    onError(
+      offset + match.index,
+      'CONTROL_CHAR',
+      `Control character \\x${hex} is not allowed and must be escaped`
+    )
+  }
+}
--- a/src/compose/resolve-block-scalar.ts
+++ b/src/compose/resolve-block-scalar.ts
@@ -3,6 +3,7 @@ import { Scalar } from '../nodes/Scalar.ts'
 import type { BlockScalar } from '../parse/cst.ts'
 import type { ComposeContext } from './compose-node.ts'
 import type { ComposeErrorHandler } from './composer.ts'
+import { checkPrintableChars } from './util-check-printable.ts'
 
 export function resolveBlockScalar(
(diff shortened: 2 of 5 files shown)


src/compose/util-check-printable.ts
```diff
@@ -0,0 +1,39 @@
...
+ * are not affected.
+ */
+export function checkPrintableChars(
+  source: string,
+  offset: number,
+  onError: ComposeErrorHandler
+): void {
+  controlCharRe.lastIndex = 0
+  let match: RegExpExecArray | null
+  while ((match = controlCharRe.exec(source)) !== null) {
```
eemeli, line 27: Why this, rather than `source.match(controlCharRe)`?

And hello, human person, this is a question for you, not your LLM.

src/compose/util-check-printable.ts
```diff
@@ -0,0 +1,39 @@
...
+const controlCharRe = /[\x00-\x08\x0b\x0c\x0e-\x1f]/g
+
+/**
+ * Reports a `CONTROL_CHAR` error for each C0 control character found in
+ * `source`, positioned at `offset + index`.
+ *
+ * The check is applied to the raw scalar source rather than its resolved
+ * value, so escape sequences in double-quoted scalars (e.g. `\t` or `\x07`)
+ * are not affected.
+ */
```
eemeli, line 19: Are these comments really adding value?

Again, this is a question for the human here.

eemeli:
Closing due to lack of a human.

PR #707, opened 2026-07-27
fix: treat out-of-range floats as strings instead of Infinity (#657)

### LINKED ISSUE #657: Can't round-trip out-of-range floating point

marceli1404:
Fixes #657

## Problem

When parsing unquoted values that "look like" floats but overflow JavaScript's `Number.MAX_VALUE`, the parser resolves them as `Infinity` or `-Infinity`, which cannot round-trip back to the original value:

```js
require("yaml").parse("gitsha: 61e9540")
// { gitsha: Infinity }  — lost the original value!
```

The YAML 1.2 spec requires that floating-point values **round-trip properly**:

> Some languages (such as Perl) provide only a "number" type that allows for both integer and floating-point values. A YAML processor may use such a type for floating-point numbers, **as long as they round-trip properly.**

## Solution

Modified the `resolve` function in `floatExp` and `float` tags (core, yaml-1.1, and JSON schemas) to check `Number.isFinite()` after `parseFloat()`. When the result is not finite (i.e., `Infinity`, `-Infinity`, or `NaN` from overflow), the original string value is returned instead, causing it to be stored as a plain string:

```js
require("yaml").parse("gitsha: 61e9540")
// { gitsha: "61e9540" }  — value preserved as string
```

**Intentional** `.inf`/`.Inf`/`.INF` and `.nan`/`.NaN`/`.NAN` syntax continue to work correctly — those are matched by the separate `floatNaN` tag which handles them explicitly.

## Changes

- `src/schema/core/float.ts`: Check `Number.isFinite()` in `floatExp.resolve` and `float.resolve`
- `src/schema/json/schema.ts`: Same check in JSON schema's float resolve
- `src/schema/yaml-1.1/float.ts`: Same check in YAML 1.1's `floatExp.resolve` and `float.resolve`
- `tests/doc/types.ts`: Updated existing overflow tests to expect string values instead of Infinity-with-source roundtrip


eemeli:
Undeclared LLM use, please read the contributing guide.

PR #709, opened 2026-08-10
fix: don't loop forever on \r, U+2028 or U+2029 at stream start

hdimer:
`lex()` never returns when the input begins with a lone `\r`, `U+2028` or `U+2029`:

```js
import { parse } from 'yaml'

parse('\u2028Z\n') // hangs forever
parse('\rZ\n')     // hangs forever
parse('\r')        // hangs forever
```

It is a busy loop rather than a stack overflow, so it pins a core and never
throws. For anything that parses YAML it does not control, that turns a stray
byte into a hang.

### Cause

`stream()` finds comments and empty lines with

```js
const emptyLineOrComment = /([ \t]*)(#.*)?$/my
```

`/m` makes `$` match before any ECMAScript LineTerminator, and that set includes
`\r`, `U+2028` and `U+2029`. YAML ends a line only at `\n` or `\r\n`, which is
exactly what `newline()` consumes. At one of the other three characters both
groups match empty and `$` succeeds, so the match is zero-length: `pos += 0`,
`newline()` consumes nothing, `stream()` returns `'stream'`, and the
`while (lexer.pos < end)` loop in `lex()` spins with `pos` never advancing.

The same mismatch affects `.` in the comment group, which also stops at
`U+2028`/`U+2029`, so a comment containing one was being cut short.

### Fix

Anchor the pattern on the line ends the lexer actually consumes:

```js
const emptyLineOrComment = /([ \t]*)(#[^\n]*?)?(?=\r?\n|$)/y
```

A match is now only possible when it is followed by `\n`, `\r\n`, or the end of
input. In the first two cases `newline()` consumes the break; the third is
unreachable while `pos < source.length`. When the pattern does not match,
`lineStart()` takes over. Either way the lexer advances, so the loop terminates.
The comment group is lazy so it runs to a real line end without eating the `\r`
of a `\r\n`.

### Scope and behaviour

This is a regression on `main`; 2.9.0 is unaffected. It arrives with the regex in
fc0a803.

I checked the change against 2.9.0 rather than trusting the suite alone. On every
input that `main` already lexes without hanging, token output is unchanged. On
the inputs that used to hang, output now matches 2.9.0. That includes the cases
where it is easy to get this subtly wrong, e.g. `parse('#a\r')` is `null` with no
document rather than a document holding a `\r` scalar.

Two deliberate omissions, in case they look like oversights:

- **Lone `\r` stays content, not a line break.** YAML 1.2 `b-break` does include a
  bare CR, so this is a real deviation, but it is pre-existing and not local to
  this regex: `newline()` only consumes `\n`/`\r\n` and `toLineEnd()` scans to
  `\n`. Making CR a break is a lexer-wide change, so I kept this patch consistent
  with the surrounding code instead. Happy to look at that separately.
- **`blockScalarHeader` and `directiveLine` have the same `/m` + `$` hazard.**
  Neither can hang, since both require a leading literal and so cannot match
  empty, but both do mis-tokenize when one of these characters appears in a
  directive or a block scalar header. I left them alone to keep this to the hang.

I found this by fuzzing `parse(stringify(v))` round-trips rather than from a
report, so there may be no issue open for it.

### Tests

Five cases in `tests/lexer.ts`: lone `\r`, `U+2028`/`U+2029` at stream start,
CRLF at stream start (which pins the `\r?`), and comments containing `U+2028` or
a lone `\r`. Expected tokens were taken from 2.9.0. Note that on a regression
these time out rather than fail, since the loop is synchronous and blocks the
worker; there is a comment above them saying so.

Full suite, `lint` and `test:types` pass.

---
Used AI assistance on this; I reviewed and tested the change myself.


eemeli:
Please review the [contributing guidelines](https://github.com/eemeli/yaml/blob/main/docs/CONTRIBUTING.md#use-of-llms-and-similar-tools) and resubmit as an issue first.

PR #711, opened 2026-08-23
fix: keep indentation indicator on empty block scalar before a comment

abhijeet117:
## Summary

For an empty block scalar followed by a document-level comment, `toString()` emitted a bare `|` header without an indentation indicator. When that output is re-parsed, the comment line is no longer outside the block scalar and becomes part of the scalar value, so `parseDocument('|5\n#comment\n').toString()` does not round-trip: the comment turns into string data. The empty-value branch now emits an indentation indicator when a top-level comment follows, matching the protection that non-empty block scalars get through forced content indentation.

## Testing

Reproduced with `parseDocument('|5\n#comment\n')`: the first `toString()` output reparsed as the value `"\n\n#comment\n"` with no comment. Added a round-trip regression test in `tests/doc/comments.ts` that fails without this change. Full vitest suite passes (3387 tests), eslint and tsc are clean.

## Checklist

- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed


eemeli:
Closing due to undeclared LLM use.

PR #714, opened 2026-09-11
Simplify line unfolding during quoted string parsing

eemeli:
Rather than trying to do everything in one regexp, it's much more performant to detect the line first, and then trim it appropriately.

This also allows us to drop the feature check for supporting old Safari versions.


eemeli:
Bah, looks like the feature check is still required.

PR #715, opened 2026-09-12
fix: error on raw C0 control characters in scalars

### LINKED ISSUE #703: No error for C0-control characters

mir-ashiq:
Fixes #703.

This adds a `NON_PRINTABLE_CHAR` error (a recoverable `YAMLParseError`) when the raw source of a scalar includes a C0 control character other than TAB, LF and CR, as those characters are excluded from the [c-printable production](https://yaml.org/spec/1.2.2/#rule-c-printable) of the YAML spec. As suggested in the issue, the check runs during composition in `resolveFlowScalar()` and `resolveBlockScalar()`, so every scalar style (plain, single-quoted, double-quoted, block literal and block folded) is covered, and each offending character is reported separately with its position.

Values that include such characters via double-quoted escape sequences (e.g. `"a\x01b"`) keep parsing without errors, as those escapes are the spec-sanctioned way to represent control characters. Error recovery is unchanged: the scalar still resolves to its source text, and only `doc.errors` gains an entry.

The other `c-printable` exclusions (DEL, C1 control characters, U+FFFE, U+FFFF) are deliberately **not** flagged, as JSON allows those characters as-is within strings and this library aims to parse valid JSON as YAML; flagging them would break e.g. the `y_string_unescaped_char_delete` and `y_string_nonCharacterInUTF-8_U+FFFF` cases in the JSON test suite. Happy to extend the check (and escape those characters in the stringifier) if full `c-printable` enforcement is preferred.

Verification:

- `npm test` — 3398 tests passed, including 11 new cases in `tests/doc/errors.ts`
- `npm run lint`
- `npx prettier --check .`
- `npm run test:types`
- `npm run build`
- `npm run test:dist` — 3398 tests passed
- `npm run test:dist:types`


eemeli:
Please review the [contributing guidelines](https://github.com/eemeli/yaml/blob/main/docs/CONTRIBUTING.md) and try again.

PR #719, opened 2026-09-13
Improve TS type ergonomics, dropping the Strict generic flag

eemeli:
The changes in #655 in particular have reduced the type variability for many parts of the document-level API, so the `Strict` generic type flag doesn't really do much any more, and is adding complexity rather than reducing it. Also, we can improve the document-level `.get()` and `.getPair()` typing to be more precise, even if TS isn't really happy in the implementation.


eemeli:
The CodeQL alert is about code that's not touched here at all, so ignoring for now.

PR #721, opened 2026-09-20
Keep indent indicator on empty root block scalar + comment

### LINKED ISSUE #687: `toString()` incorrectly converts comment to string

dyk1454683243-sudo:
### Issue

Fixes #687

### LLM Use

An assistant helped apply the one-line header change and add the regression test. The wording here is mine. No generated comments in the code.

### Description

`toString()` wrote an empty root `|` / `>` without an indent indicator, so a following document comment parsed as the scalar.

When `forceBlockIndent` is set, emit `|2` / `>2` so the comment stays a comment. Nested empty scalars already round-trip; this only covers the root case.

#697 tried a wider version of this and was closed by the author after review.


eemeli:
Please quote to me here the last sentence of the contributing guide, which I trust you've read?

